PR #4, opened 2024-07-12
Map function types to names and update output format

sskeirik:
This PR makes a series of changes:

1.  the biggest one --- a new top-level field `functions` is added that points to a map from interned `Ty`s (numbers) to link-time function symbol names (strings)
2.  the structure of `MonoItemFn`s changed as follows:
    -   several debug fields were removed and put in the `ItemDetails` struct instead
    -   the `body` and `promoted` array were unified into a single array field
3.  the `MirBody` struct was removed and the `BodyDetails` struct was moved to a new array field in `ItemDetails`
4.  some new env vars were added to control how the link map is built and how debug information is emitted

The construction of the linking map is done in two phases:

-   (optionally) each `MonoItem::Fn(Instance)` has its `Instance`s underlying function type mapped to the `Instance`s link-time mangled name
-   for each `MonoItem::Fn(Instance)`, we walk over its body and any promoted bodies, find each `Call` terminator/`ReifyFnPointer` cast, and then map the corresponding `Instance` type for the `Call` or `fn` pointer cast to its link-time mangled `Instance` name

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ To generate stable MIR output without building a binary, you can invoke the tool
 ./run.sh -Z no-codegen <crate_root>
 ```
 
+There are a few environment variables that can be set to control the tools output:
+
+1.  `LINK_ITEMS` - add entries to the link-time `functions` map for each monomorphic item in the crate;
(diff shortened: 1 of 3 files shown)


mariaKt (approved):
Approved. @dkcumming @sskeirik

PR #5, opened 2024-07-22
Improve interned id printing

sskeirik:
This PR helps with interned id processing and makes three changes:

1. we now grab and serialize the stable crate id from the compiler for use in later processing
2. we re-order the crate field serialization so that the alloc id and function maps are printed first
3. the example panic file is updated


mariaKt (approved):
The output is as expected, thanks.

PR #18, opened 2024-11-01
Remove imported kani code

sskeirik:
For the moment, we will remove the kani monomorphization logic, since that code was less well-understood than our custom code that hooked into the compiler monomorphization routines directly.

This also fixes a build error when moving to depend on rustc master (as of today's date).


dkcumming (approved):

dkcumming:
Thank you!

PR #21, opened 2024-11-12
Initial CI setup and a few tests

jberthold:
* [X] Simple CI setup to build and test smir_pretty
* [x] a directory of small test programs and golden files to compare to (json to be formatted and normalised using `jq`) 
* [x] resolve issues with toolchain setup (from prior runs?) on runners
* [ ] analyse and fix Rust build artefact caching

Related: #20


tests/integration/failing/array.rs
jberthold, line 1: Most files in this directory fail the test because their `functions` array content is not deterministic, the integers paired with the function names are different in different runs.

tests/integration/failing/panic_example.rs
jberthold, line 1: This test fails because of a `Static` element in the `allocs`. None of the other tests shows this `Static` element.

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+          make build_all
+
+      - name: 'Run smir integration tests'
+        run: |
+          make integration-test
+
+      - name: 'Clean up toolchain'
+        if: always()
+        run: |
+          make rustup-clear-toolchain
```
jberthold, line 60: This is important, otherwise the build will leave a dangling toolchain behind that `rustup` stumbles upon during setup. I [opened a ticket for this against `dtolnay/rust-toolchain@`](https://github.com/dtolnay/rust-toolchain/issues/127).

dkcumming: Looks like a solution is already merged @jberthold

dkcumming (approved):
Okay it all looks sensible to me. I think this is great for merge

PR #24, opened 2025-01-15
Filter and normalise json output for golden test runs

dkcumming:
This PR adds functionality to filter and normalise the json output when comparing output to the golden tests. Some cases when we want to normalise output:
1. `"NormalSym"` mangled function names will have a hash following `17h` that will [change when the compiler version changes](https://github.com/rust-lang/rust/blob/2ae9916816a448fcaab3b2da461de754eda0055a/compiler/rustc_symbol_mangling/src/lib.rs#L69-L76). 
```
<         "NormalSym": "_ZN3std2rt19lang_start_internal17h3166ab4c34d1da61E"
---
>         "NormalSym": "_ZN3std2rt19lang_start_internal17h51b943990c0e2c96E"
                                                        ^----can ignore---^
```
2. `"symbol_name"` mangled function names will have a hash following `17h` that will [change when the compiler version changes](https://github.com/rust-lang/rust/blob/2ae9916816a448fcaab3b2da461de754eda0055a/compiler/rustc_symbol_mangling/src/lib.rs#L69-L76). 
```
<       "symbol_name": "_ZN4core3ops8function6FnOnce9call_once17h40ba616aaa505a69E"
---
>       "symbol_name": "_ZN4core3ops8function6FnOnce9call_once17hd98db8d17b807164E"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,8 @@ TESTDIR=$(CURDIR)/tests/integration/programs
 .PHONY: integration-test
 integration-test: TESTS     ?= $(shell find $(TESTDIR) -type f -name "*.rs")
 integration-test: SMIR      ?= $(CURDIR)/run.sh -Z no-codegen
+# before comparison we want to clean the files up
+integration-test: PRE_FILTER ?= jq -e -f $(TESTDIR)/../pre-filter.jq
 # override this to tweak how expectations are formatted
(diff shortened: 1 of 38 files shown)


jberthold (approved):
LGTM
Have to check what happened to the test workflow...

PR #25, opened 2025-01-15
Run test workflow on (public) ubuntu-latest runner

jberthold:


dkcumming:
@jberthold can we delete this?

jberthold:
> @jberthold can we delete this?

Yes, probably not too important anyway now that we have short builds.

PR #26, opened 2025-01-15
Changed the expected files to remove ids

dkcumming:


dkcumming:
Discussed with Jost, we think it is good to merge

PR #30, opened 2025-01-21
Temporary fix for layout of panic strings

dkcumming:
Creating the layout for a `Binder<DUMMY>` is creating panics, see https://github.com/runtimeverification/smir_pretty/issues/27 for details. This PR does not solve the problem but side steps it. There may be other times that this will be needed unless we find a fix to get the `Layout` without triggering the error.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -473,8 +473,33 @@ fn collect_arg_tys(collector: &mut InternedValueCollector, args: &stable_mir::ty
 }
 
 fn collect_ty(val_collector: &mut InternedValueCollector, val: stable_mir::ty::Ty) {
-    use stable_mir::ty::{RigidTy::*, TyKind::RigidTy}; // GenericArgKind::*, TyConst, TyConstKind::*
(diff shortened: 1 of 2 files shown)


src/printer.rs
```diff
@@ -473,8 +473,33 @@ fn collect_arg_tys(collector: &mut InternedValueCollector, args: &stable_mir::ty
 }
 
 fn collect_ty(val_collector: &mut InternedValueCollector, val: stable_mir::ty::Ty) {
-    use stable_mir::ty::{RigidTy::*, TyKind::RigidTy}; // GenericArgKind::*, TyConst, TyConstKind::*
-    if val_collector.visited_tys.insert(hash(val), (val.kind(), val.layout().map(|l| l.shape()).ok())).is_some() {
+    use stable_mir::ty::{RigidTy::*, TyKind::RigidTy, AdtDef}; // GenericArgKind::*, TyConst, TyConstKind::*
```
jberthold, line 476: Maybe remove the old commented code while we are changing this line anyway... 
```suggestion
    use stable_mir::ty::{RigidTy::*, TyKind::RigidTy, AdtDef};
```

jberthold (approved):

PR #33, opened 2025-01-24
Removing compiler build and using `rustup`

dkcumming:
Changing to use `rustup` for the Rust compiler now that we are not on our own fork.

- [x] Need `build.rs` so that correct rust version is used with components installed
- [x] Need to update README.md
- [x] Building with `Makefile` needs to be removed
- [x] Handling tests with `Makefile` needs to change to use `cargo`
- [x] CI needs to be updated succesfully
- [x] Need to handle golden tests difference
- [x] Need to make sure this will still work with `mir-semantics` ([see this PR](https://github.com/runtimeverification/mir-semantics/pull/439))

Moved to future work:
- [ ] Change tests to use `cargo test` ? Could remove `Makefile` entirely then
- [ ] Need to make sure this will still work with extracting stable-mir-json for cargo projects / `std lib`


Makefile
```diff
@@ -1,115 +1,13 @@
...
-
-setup: rust_clone
-
-update: ${RUST_SRC}
-	cd "${RUST_SRC}"; git fetch origin; git checkout ${RUST_COMMIT}
-
-build:
-	cargo build ${RELEASE_FLAG}
-
-clean:
```
jberthold, line 44: Should we maybe keep `build`, and update `clean` to do the right thing?

Makefile
```diff
@@ -1,115 +1,13 @@
...
-	rustup override unset
-	rustup override unset --nonexistent
-	rustup toolchain uninstall "${TOOLCHAIN_NAME}"
-
-generate_ui_tests:
-	mkdir -p "${RUST_DIR}"/tests
-	cd "${RUST_SRC}"; ./get_runpass.sh tests/ui > "${RUST_DIR}"/tests_ui_sources
-	-cd "${RUST_SRC}"; ./ui_compiletest.sh "${RUST_SRC}" "${RUST_DIR}"/tests/ui/upstream "${RUST_DIR}"/tests_ui_sources --pass check --force-rerun 2>&1 > "${RUST_DIR}"/tests_ui_upstream.log
-	-cd "${RUST_SRC}"; RUST_BIN="${PWD}"/run.sh ./ui_compiletest.sh "${RUST_SRC}" "${RUST_DIR}"/tests/ui/smir "${RUST_DIR}"/tests_ui_sources --pass check --force-rerun 2>&1 > "${RUST_DIR}"/tests_ui_smir.log
-
```
jberthold, line 102: ...reminds me I wanted to still check what exactly this is doing - more test programs and maybe more failure modes that we haven't seen... but we can still replicate this later.

jberthold (approved):
LGTM, tested locally as well

Makefile
dkcumming: I think it is still accessible on our fork also could add an issue as a reminder there and backlog it

PR #47, opened 2025-02-13
Working with cargo

### LINKED ISSUE #39: Support calling smir_pretty from inside another cargo project
Following #33 we have issues calling smir_pretty from inside another cargo project. This is because smir_pretty now runs off  cargo itself and the nested cargo problem needs to be understood.

This is needed for two reasons:
1. Emitting the stable-mir-json for the standard library
2. Emitting all the stable-mir-json for the source of a cargo project

dkcumming:
This didn't end up looking anything like what I had imagined. But I was having a lot of trouble getting this working at all and I may have had my environment set up in a way that was self defeating as I feel I suddenly started having things work when trying the same thing. Not sure.

But right now this is creating a `.stable_mir_json` directory at a given directory or `~/.stable_mir_json` directory (linux only right now) and that is populated with the binary for `stable_mir_json`,  a file containing the `LD_LIBRARY_PATH` that was used to create the binary, and scripts to set the library path so that RUSTC can be overwritten when calling cargo. 

To test this out:
1. Fetch and switch to this branch
2. run `cargo build` and `cargo build --release`
3. run `cargo run --bin cargo_stable_mir_json -- $PWD [optional-dir-path]`  from the root of the repo (this creates `.stable_mir_json`)
4. create or navigate to another cargo project you want to test
5. run `RUSTC=~/.stable_mir_json/debug.sh cargo build` (or the file that `.stable_mir_json` was built in). Or for release `RUSTC=~/.stable_mir_json/release.sh cargo build`

Fixes #39


Cargo.toml
```diff
@@ -1,13 +1,19 @@
 [package]
-name = "stable_mir_pretty"
+name = "stable_mir_json"
```
jberthold, line 2: :man_facepalming: sorry for that

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+fn record_ld_library_path(hidden_dir: &Path) -> Result<()> {
+    const LOADER_PATH: &str = "LD_LIBRARY_PATH";
+    if let Some(paths) = env::var_os(LOADER_PATH) {
+        // Note: kani filters the LD_LIBRARY_PATH, not sure why as it is working locally as is
+        let mut ld_library_file = std::fs::File::create(hidden_dir.join("ld_library_path"))?;
+        writeln!(ld_library_file, "{}", paths.to_str().unwrap())?;
+    } else {
+        bail!("Couldn't read LD_LIBRARY_PATH from env");
```
jberthold, line 54: Do we know for sure this cannot be empty? 
If yes, then maybe reformulate the error message to say so, and why.
If no, we should not `bail`.

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,181 @@
...
+fn add_run_script(hidden_dir: &Path) -> Result<()> {
+    let run_script_path = hidden_dir.join("run.sh");
+    let mut run_script = std::fs::File::create(&run_script_path)?;
+    writeln!(run_script, "#!/bin/bash")?;
+    writeln!(run_script, "set -eu")?;
+    writeln!(run_script)?;
+    writeln!(
+        run_script,
+        "export LD_LIBRARY_PATH=$(cat ~/.stable_mir_json/ld_library_path)"
+    )?;
```
jberthold, line 36: Maybe we can just pass the `ld_library_path` contents to this function as a string ?
Also, the hard-coded `.stable_mir_json` is probably `hidden_dir`, right?

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,181 @@
...
+            Profile::Release => "release/".into(),
+        }
+    }
+}
+
+fn cp_artefacts_from_profile(
+    profile_dir: PathBuf,
+    smir_json_dir: &Path,
+    profile: Profile,
+) -> Result<()> {
```
jberthold, line 128: This code seems to double up the distinctin between `debug/` and `release./` in both the `profile_dir` and in the `profile` variable and related `impl`.

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,181 @@
...
+    writeln!(run_script, "#!/bin/bash")?;
+    writeln!(run_script, "set -eu")?;
+    writeln!(run_script)?;
+    writeln!(
+        run_script,
+        "export LD_LIBRARY_PATH=$(cat ~/.stable_mir_json/ld_library_path)"
+    )?;
+    writeln!(
+        run_script,
+        "exec \"/home/daniel/.stable_mir_json/debug/stable_mir_json\" \"$@\""
```
jberthold, line 39: Maybe use `$HOME` here?
```suggestion
        "exec \"$HOME/.stable_mir_json/debug/stable_mir_json\" \"$@\""
```

jberthold (commented):
Comments are a few nit-picks on details...
The approach as a whole seems reasonable for distribution - maybe there would be a more canonical way to place things inside `$HOME/.cargo` but if this works we can go with this for now.

Can you add a usage example here, and a description in the `README.md`?

jberthold:
Might be good to give users the option of a custom install path.

src/bin/cargo_stable_mir_json.rs
dkcumming: I think that this should be unreachable, I can't see that the `cargo_stable_mir_json` could run without `LD_LIBRARY_PATH` being set. But maybe I am missing something, when do you think this could fail?

dkcumming: Oh I just realised that means I am in the "If, yes" case of what you said. I will change the error message. I don't think there is an `anyhow::unreachable` but if there is I will use that too.

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,178 @@
...
+    writeln!(run_script, "set -eu")?;
+    writeln!(run_script)?;
+    writeln!(
+        run_script,
+        "export LD_LIBRARY_PATH={}",
+        ld_library_path.display(),
+    )?;
+    writeln!(
+        run_script,
+        "exec \"{}/debug/stable_mir_json\" \"$@\"",
```
dkcumming, line 153: Need to have some arg for choosing `release` instead of `debug`

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+fn setup(repo_dir: PathBuf, maybe_user_provided_dir: Option<PathBuf>) -> Result<()> {
+    let smir_json_dir = smir_json_dir(maybe_user_provided_dir)?;
+    println!("Creating {} directory", smir_json_dir.display());
+    std::fs::create_dir(&smir_json_dir)?; // This errors is the directory already exists
+
+    let ld_library_path = record_ld_library_path(&smir_json_dir)?;
+    copy_artefacts(&repo_dir, &smir_json_dir, &ld_library_path)
+}
```
jberthold, line 44: Maybe print a "success" message here?
```suggestion
    copy_artefacts(&repo_dir, &smir_json_dir, &ld_library_path)
    println!("Artefacts installed in {}.", &simr_json_dir);
    println!("To use stable-mir-json, set `RUSTC={}/\{debug,release\}/stable-mir-json` when calling `cargo`.", smir_json_dir);
}
```

src/bin/cargo_stable_mir_json.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    setup(repo_dir, maybe_user_provided_dir)
+}
+
+fn setup(repo_dir: PathBuf, maybe_user_provided_dir: Option<PathBuf>) -> Result<()> {
+    let smir_json_dir = smir_json_dir(maybe_user_provided_dir)?;
+    println!("Creating {} directory", smir_json_dir.display());
+    std::fs::create_dir(&smir_json_dir)?; // This errors is the directory already exists
```
jberthold, line 40: `This errors is the directory already exists` ... not sure that's desirable. We can leave it that way for now, though...
Also, when I suggested the user-defined location I thought this would be the final destination, not that a directory would be created there.

jberthold (approved):
Sorry, I still have a few comments on details... and should we maybe add a smoke test in CI that runs this in docker and tries to use the resulting installation?

We can also merge this and refactor on a follow-up PR to make progress

dkcumming:
Added the other issue here https://github.com/runtimeverification/stable-mir-json/issues/48

Issue #32, opened 2025-01-23
Need to handle provenance for all `RigidTy`s


dkcumming:
We have unfinished work dumping the provenance for `RigidTy`. We need to complete the variants in [this match statement](https://github.com/runtimeverification/smir_pretty/blob/5314e1f8d8053dc129313741bbd3e77ddd1cf0c9/src/printer.rs#L424).

dkcumming:
#49 Will implement some variants

Issue #27, opened 2025-01-16
Compiler panic on programs using `to_string()`


jberthold:
<summary>
When extracting smir json for a program that contains a call to `to_string()`, the `printer.rs` code crashes with

```
thread 'rustc' panicked at compiler/rustc_middle/src/ty/predicate.rs:533:9:
`<[core::fmt::rt::Argument<'_>] as std::marker::Sized>` has escaping bound vars, so it cannot be wrapped in a dummy binder.
```

Programs that exhibit this are `tests/integration/failing/{defined-trait,std-to-string}.rs`

The backtrace is here:

<details>

```
$ RUST_BACKTRACE=1 deps/smir_pretty/run.sh -Z no-codegen kmir/src/tests/integration/data/run-rs/strings-chars/std-to-string.rs 

thread 'rustc' panicked at compiler/rustc_middle/src/ty/predicate.rs:533:9:
`<[core::fmt::rt::Argument<'_>] as std::marker::Sized>` has escaping bound vars, so it cannot be wrapped in a dummy binder.
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <rustc_middle::ty::predicate::Predicate as rustc_type_ir::upcast::UpcastFrom<rustc_middle::ty::context::TyCtxt, rustc_type_ir::predicate::TraitRef<rustc_middle::ty::context::TyCtxt>>>::upcast_from
   3: rustc_trait_selection::traits::type_known_to_meet_bound_modulo_regions
   4: rustc_ty_utils::common_traits::is_item_raw
      [... omitted 2 frames ...]
   5: rustc_middle::query::plumbing::query_get_at::<rustc_query_system::query::caches::DefaultCache<rustc_middle::ty::PseudoCanonicalInput<rustc_middle::ty::Ty>, rustc_middle::query::erase::Erased<[u8; 1]>>>
   6: <rustc_middle::ty::Ty>::is_sized
   7: rustc_ty_utils::layout::layout_of
      [... omitted 2 frames ...]
   8: <rustc_middle::ty::layout::LayoutCx as rustc_middle::ty::layout::LayoutOf>::spanned_layout_of
[...]

PR #49, opened 2025-02-19
Fixing some errors so p-token builds

dkcumming:
Currently `stable-mir-json` will not build complex programs. This PR makes some changes so that [p-token](https://github.com/febo/p-token) will build as long as [this feature flag in pinocchio](https://github.com/anza-xyz/pinocchio/blob/1a10e2277303d843e8f2922d4efab87f41d82aff/sdk/log/crate/Cargo.toml#L20) is removed (this is unneeded to build p-token so it is no problem to remove it)

- Added `Adt` and `Ref` type handling for the `RigidTy`s in `get_prov_type` as seen in #32 
- Side steps the DUMMY_BINDER escaping bound vars in formatting arguments as seen in #27


jberthold (approved):
LGTM :rocket:

PR #51, opened 2025-02-21
Render projections in the SMIR graph of basic blocks

jberthold:
* field projections are rendered as suffix `.N` where `N` is the field index
* array indexing (constant or with a local) is rendered as `[IDX]`, where `IDX` is either a local written with `_` prefix or a plain integer
* dereferencing is rendered as a `*` suffix (NB _not_ prefix, for practicality)
* type casts and subtyping are rendered by appending `as _TYPE_`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,3 +42,9 @@ integration-test:
 
 golden:
 	make integration-test DIFF=">"
+
+format: 
+	cargo fmt
+
+style-check: format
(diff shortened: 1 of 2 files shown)


dkcumming (approved):
Looks good. I think the `*`  at the suffix doesn't matter for now. We will iron things out over time.

PR #56, opened 2025-03-13
Small tweaks to graph and installed script

jberthold:
Tweaks to graph rendering
* Reference creation and dereferencing are now rendered differently:
   - dereference: `& ...` or `&mut ...`, 
* operand copy and operand move are now indicated by `cp(...)` and `mv(...)`

Tweaks to installed run script:
* environment variable `STABLE_MIR_OPTS` can be set to pass `--dot` or `--json` (or any number of other options) directly to the compiler invocation
* the script now uses its location as a path prefix instead of fixing it to the installation directory, i.e., the installation can be moved to a different directory without breaking

### DIFF
--- a/src/mk_graph.rs
+++ b/src/mk_graph.rs
@@ -15,8 +15,8 @@ use rustc_session::config::{OutFileName, OutputType};
 
 extern crate rustc_session;
 use stable_mir::mir::{
-    AggregateKind, BasicBlock, ConstOperand, Mutability, NonDivergingIntrinsic, NullOp, Operand,
-    Place, ProjectionElem, Rvalue, Statement, StatementKind, TerminatorKind, UnwindAction,
+    AggregateKind, BasicBlock, BorrowKind, ConstOperand, Mutability, NonDivergingIntrinsic, NullOp,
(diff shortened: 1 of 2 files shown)


src/mk_graph.rs
```diff
@@ -397,54 +397,58 @@ impl GraphLabelString for Operand {
...
+            format!(
+                "{thing}[{}..{}{}]",
+                from,
+                if *from_end { "-" } else { "" },
+                to
+            )
         }
+        ProjectionElem::Downcast(i) => format!("{thing} as {:?}", i),
+        ProjectionElem::OpaqueCast(ty) => format!("{thing} as type {}", ty),
+        ProjectionElem::Subtype(i) => format!("{thing} as {:?}", i),
```
dkcumming, line 436: Should we add something like `as subtype` just so it's clear it isn't a downcast?

dkcumming (approved):
All looks great

src/mk_graph.rs
jberthold: Might be good, yes. I should check which projections are actually used in `p-token`, though...

PR #60, opened 2025-03-24
Removed the multiple `Body`s for `Instance`

dkcumming:
Stable MIR `Instance` `Body` has already allocated promoted, [see this conversation on zulip]([#project-stable-mir > Getting fully monomorphised &#96;MonoItem&#96;s @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/320896-project-stable-mir/topic/Getting.20fully.20monomorphised.20.60MonoItem.60s/near/507085029))

### DIFF
--- a/src/mk_graph.rs
+++ b/src/mk_graph.rs
@@ -223,7 +223,7 @@ impl SmirJson<'_> {
 
                         if let Some(body) = &body {
                             process_blocks(&mut c, 0, &body.blocks);
-                        } // TODO: Should we indicate missing `Body`?
+                        } // TODO: Investigate `MonoItem::Fn` without `Body` and make dummy node if correct
 
                         drop(c); // so we can borrow graph again
 
@@ -280,7 +280,7 @@ impl SmirJson<'_> {
 
                         if let Some(ref body) = body {
                             add_call_edges(&mut graph, 0, &body.blocks);
-                        } // TODO: Should we indicate missing `Body`?
+                        }
                     }
                     MonoItemKind::MonoItemGlobalAsm { asm } => {
                         let mut n = graph.node_named(short_name(&asm));
--- a/src/mk_graph.rs
+++ b/src/mk_graph.rs
@@ -221,7 +221,10 @@ impl SmirJson<'_> {
                                 }
                             };
 
-                        process_blocks(&mut c, 0, &body.blocks);
+                        if let Some(body) = &body {
+                            process_blocks(&mut c, 0, &body.blocks);
+                        } // TODO: Should we indicate missing `Body`?
+
                         drop(c); // so we can borrow graph again
 
                         // call edges have to be added _outside_ the cluster of blocks for one function
@@ -275,7 +278,9 @@ impl SmirJson<'_> {
                                 }
                             };
 
-                        add_call_edges(&mut graph, 0, &body.blocks);
+                        if let Some(ref body) = body {
+                            add_call_edges(&mut graph, 0, &body.blocks);
+                        } // TODO: Should we indicate missing `Body`?
                     }
                     MonoItemKind::MonoItemGlobalAsm { asm } => {
                         let mut n = graph.node_named(short_name(&asm));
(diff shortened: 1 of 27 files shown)


dkcumming:
I thought I could ensure that every `Instance` from `MonoItem::Fn` and `MonoItem::Static` would have a `Body`, but there are times when this does not seem possible. Related to [this dump for extra `Static`s](https://github.com/runtimeverification/stable-mir-json/pull/60/files#diff-64932fd51f70fd457567303b06d7e4e92efb03febb303d5a1e0ba7ac82d4014dR944-R958) from @sskeirik if we attempt to allocate memory on the heap with something like `Box::new`,  then [this `Static` related to global allocation `__rust_no_alloc_shim_is_unstable`](https://github.com/rust-lang/rust/blob/ae8ab87de4d8caab5d91a027bc19bb5d5e8a3691/library/alloc/src/alloc.rs#L35) will not have a `Body`.

By following the links on [this Zulip comment]([#general > Potential rustc compiler issue with too many Boxes ? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Potential.20rustc.20compiler.20issue.20with.20too.20many.20Boxes.20.3F/near/500351059)) it looks like it is about supporting multiple global allocators and ensuring that some reads don't get optimised out when they shouldn't

src/mk_graph.rs
```diff
@@ -221,22 +221,10 @@ impl SmirJson<'_> {
...
-                                for b in &body {
-                                    let mut cc = c.cluster();
-                                    process_blocks(&mut cc, curr, &b.blocks);
-                                    curr += b.blocks.len();
-                                }
-                            }
-                        }
+                        if let Some(body) = &body {
+                            process_blocks(&mut c, 0, &body.blocks);
+                        } // TODO: Should we indicate missing `Body`?
```
jberthold, line 226: I guess we need at least one dummy node in the cluster so we know it exists.

src/mk_graph.rs
```diff
@@ -290,19 +278,9 @@ impl SmirJson<'_> {
...
-                                let mut curr: usize = 0;
-                                for b in &body {
-                                    add_call_edges(&mut graph, curr, &b.blocks);
-                                    curr += b.blocks.len();
-                                }
-                            }
-                        }
+                        if let Some(ref body) = body {
+                            add_call_edges(&mut graph, 0, &body.blocks);
+                        } // TODO: Should we indicate missing `Body`?
```
jberthold, line 283: Nothing to do here, this code is for _outgoing_ edges

src/printer.rs
```diff
@@ -771,31 +746,39 @@ fn collect_interned_values<'tcx>(tcx: TyCtxt<'tcx>, items: Vec<&MonoItem>) -> In
...
-                        _sym: inst.mangled_name(),
-                        locals: &[],
-                        link_map: &mut calls_map,
-                        visited_tys: &mut visited_tys,
-                        visited_allocs: &mut visited_allocs,
-                    }
-                    .visit_body(&body)
+                let body = inst.body().unwrap_or_else(|| {
+                    panic!(
+                        "Failed to retrive body for Instance of MonoItem::Fn {}",
```
jberthold, line 769:
```suggestion
                        "Failed to retrive body for Instance of MonoItem::Static {}",
```

src/printer.rs
```diff
@@ -771,31 +746,39 @@ fn collect_interned_values<'tcx>(tcx: TyCtxt<'tcx>, items: Vec<&MonoItem>) -> In
...
+                        inst.name()
+                    )
+                });
```
jberthold, line 754: So this looks like the cases you found where `MonoItem:Fn` does not have a body are somehow filtered out before? Or would we rather want to just report it (or not even that) and move on when there is no `body` to visit?

jberthold (approved):
Not entirely clear why we panic on body-less functions but we can merge and amend if needed (required downstream changes should be added quickly)

src/mk_graph.rs
dkcumming: Actually I am going to make an issue for this, while going to fix it I now am confused if `__rustc_no_alloc_shim_is_unstable` is a `Static` or a `Fn`, and I wonder if there is some little error to clear up there. I will make an issue for it.

PR #65, opened 2025-04-05
Updates: Graph and `Adt` names

dkcumming:
- I removed rendering an empty node for the graph if the `MonoItem::Fn` has no `Body`, I have not had time to dig much more into it but the graph should have the empty node replaced. 

- Accessing the name for `Adt` is not done with the `Stable MIR` API, updating to doing this means the names are slightly different (as can be seen in the diff) but I don't think the information using the internal method has anything needed

### DIFF
--- a/src/mk_graph.rs
+++ b/src/mk_graph.rs
@@ -223,7 +223,9 @@ impl SmirJson<'_> {
 
                         if let Some(body) = &body {
                             process_blocks(&mut c, 0, &body.blocks);
-                        } // TODO: Investigate `MonoItem::Fn` without `Body` and make dummy node if correct
+                        } else {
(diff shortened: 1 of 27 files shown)


src/mk_graph.rs
```diff
@@ -223,7 +223,9 @@ impl SmirJson<'_> {
 
                         if let Some(body) = &body {
                             process_blocks(&mut c, 0, &body.blocks);
-                        } // TODO: Investigate `MonoItem::Fn` without `Body` and make dummy node if correct
+                        } else {
+                            c.node_auto().set_label("<empty body>");
```
jberthold, line 227: :+1:

jberthold (approved):

PR #68, opened 2025-04-22
Use a type visitor to collect type metadata, remove layout

### LINKED ISSUE #67: Collect type metadata using a type table visitor instead of searching items
Types used in `TyConstKind::Value` are apparently not collected by the visitor that collects type information to emit in stable-mir-json. 
We should instead use a type table visitor (to be investigated).

jberthold:
Fixes #67 

As discussed, using a type visitor to collect type information is more robust than the prior `collect_ty`.
Layout information causes issues with polymorphic functions again, so we do not collect it any more for now as we aren't currently using it at all in `mir-semantics`.


src/printer.rs
```diff
@@ -477,13 +478,33 @@ type AllocMap = HashMap<stable_mir::mir::alloc::AllocId, AllocInfo>;
...
+        TyCollector {
+            types: HashMap::new(),
+        }
+    }
+}
+
+impl Visitor for TyCollector {
+    type Break = ();
+    fn visit_ty(&mut self, ty: &stable_mir::ty::Ty) -> ControlFlow<Self::Break> {
+        self.types.insert(*ty, (ty.kind(), None));
```
dkcumming, line 496: Actually thinking about this again, I don't think this is a good idea, because the attempt to get a `Layout` failing is because we are still collecting `Ty`s on unmonomorphised function like items. I think the vistor / collector needs to be corrected before we merge this on second thoughts.

jberthold: Agree this is indicative of a problem we have (quite probably in stdlib code only, but still).
We can maybe also just fix the `collect_ty` function to consider types within `TyConst` (that was the original motivation to look into this).

gtrepta: I updated the visitor to resolve the unmonomorphised items like before.

Now, this visitor does a way better job of visiting every possible thing in these types. This is good, obviously, but it also means we hit some stuff we weren't previously that we don't know how to handle just yet. So, I've arranged the visitor to break on these items to avoid any panics. These are:

- Bound regions (`RegionKind::ReBound`). Some of these have escaping vars that cause a message saying they can't be wrapped in a dummy binder. There may be some bound regions that don't have this issue, but at the moment I'm just skipping all of them.
- `RigidTy::CoroutineWitness`. Trying get the layout for these causes a panic. Maybe we just skip calling layout and use `None`, but I have it skipping them altogether right now.

This passes the ui tests now. I think this is in a good place now and should make it much easier to special case specific things and track what we aren't supporting.

tests/integration/programs/assert_eq.smir.json.expected
jberthold, line 1: I am a bit puzzled by the differences in the expectation files.
The code does not collect any function types any more ? That's OK I guess, we don't use them, but... we also don't get `I8` any more (in this example)? Why not?

jberthold:
Some IDs seem to be unstable (type info records change location in the expectation files on different runs). We could sort the type infos by something that stabilises the order (the `name` is the obvious option for the failure I saw in CI - `jq` should have a stable `sort` function so things without name don't change order).

src/printer.rs
dkcumming: Is this intended as a temporary fix? I am hesitant to move forward merging in code knowing that we are accessing MIR that is not resolved. But I understand that if we don't know where it is originating from and we need the `Ty`s collected immediately we need to weigh that in too.

dkcumming: Or am I misunderstanding what we are avoiding with breaking the control flow?

tests/integration/programs/assert_eq.smir.json.expected
dkcumming: I also can't understand why `i8` is now removed from the map

src/printer.rs
gtrepta: The two bullet points I listed above are cases where we _don't_ add the type to the metadata output. It's intended to prevent any panics at runtime, and it's temporary in the sense that we (I) don't understand why the panic happens or how to resolve it.

As far as whether or not we're still hitting some MIR, maybe we can determine somehow if we are with breakpoints/cases/print statements. Not sure.

dkcumming (approved):
I think great improvement, you mentioned some more things to do but that can be another PR

PR #70, opened 2025-04-25
Remove `FunType` from `TypeInfo` metadata

jberthold:
Rendering the function types as strings caused problems downstream when writing K claim files so we remove the output for now. We might change to a different type representation for function types anyway.


gtrepta (approved):

jberthold:
Obsolete after solving the underlying parser problem, closing

PR #71, opened 2025-04-26
Handling `FnPtr` signature internally to avoid `Binder`s

dkcumming:
I think this is the correct way to access the function signature args for a function pointer, even though it goes through internal. I checked kani kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs and this is what they are doing. 

There was one failure `rust/tests/ui/issues/issue-26641.rs (exit 101)` that uses a `RigidTy::Dynamic(Binder...)` but we haven't really looked at details for `dyn` and `vtable` yet so this is moved to failing for now.

Also improved the UI testing output but adding verbosity and some metrics


src/printer.rs
```diff
@@ -488,21 +487,23 @@ type AllocMap = HashMap<stable_mir::mir::alloc::AllocId, AllocInfo>;
 type TyMap =
     HashMap<stable_mir::ty::Ty, (stable_mir::ty::TyKind, Option<stable_mir::abi::LayoutShape>)>;
 
-struct TyCollector {
+struct TyCollector<'tcx> {
+    tcx: TyCtxt<'tcx>,
```
jberthold, line 491: :+1: makes a lot of sense to carry the `tcx`

tests/ui/failing.tsv
jberthold, line 1: This looks like very good progress - there don't seem to be any `134` exit codes left (? haven't checked thoroughly, just eye-balled it)

tests/ui/run_ui_tests.sh
```diff
@@ -26,18 +28,31 @@ while read -r test; do
...
     status=$?
 
+    total=$((total + 1))
+
     if [ $status -ne 0 ]; then
         echo "❌ FAILED: $test_path (exit $status)"
-        failed=1
+        failed=$((failed + 1))
+    else
+        # echo "✅ PASSING: $test_path"
```
jberthold, line 37: We should probably just have a verbosity option here... which is of course the most important part of this change :smirk:

jberthold (approved):
Code makes sense and appears to be a definite improvement.
Happy to have it merged (to master on top of Guy's branch), or we leave it for Guy to chip in otherwise.

tests/ui/failing.tsv
dkcumming: I was thinking the same thing, I think 134 was recursion limit right?

tests/ui/run_ui_tests.sh
dkcumming: Sure I commented it out when I was trying to just get a list of the new failing tests, but I put it with intentions to turn it on. I will change now

PR #72, opened 2025-04-29
Graph improvements

dkcumming:
Some readability issues are fixed in this PR:

- `Deref` projections are now wrapped in parenthesis e.g.`(*4)`
- Constants are not longer separated by "`::`", e.g. `const :: u32` is now `const ?_u32`, and `const :: std::ops::ExitCode` is now `const std::ops::ExitCode`. I do not like that there is "`?`" but I couldn't see a fast way to deal with the bytes in the `Allocation`, does anyone know how to parse that into the integer?
- Changed `Rvalue::AddressOf` e.g. `&Mut (*3)` is now `&raw mut (*3)`, which is less likely to be confused with a mutable borrow `&mut (*3)`
- Added a node with all the locals and type information. Some formatting and placement issues with this. I would appreciate some feedback!


gtrepta (approved):
LGTM.

The `MirConst` from `const ?_u32` appears to have a `eval_target_usize` method which might return the constant as a u64. I tried to add it in real quick but got a compiler panic, `"expected int of size 8, but got size 4"`. It might be pretty sensitive to how it gets called, but I didn't mess with it any further.

src/mk_graph.rs
```diff
@@ -87,6 +91,16 @@ impl SmirJson<'_> {
...
                         }
 
+                        // Set out the type information of the locals
+                        let mut local_node = c.node_auto();
+                        let mut vector: Vec<String> = vec![];
+                        vector.push(String::from("LOCALS"));
+                        for (index, decl) in body.clone().unwrap().local_decls() {
+                            vector.push(format!("{index} = {}", decl.ty));
+                        }
+                        local_node.set_label(vector.join("\\l").to_string().as_str());
```
jberthold, line 101: Maybe add an empty string at the end to get the last line with `\l`, too? 
And we could maybe set a different colour to contrast it with normal nodes.
```suggestion
                        vector.push(String::from("LOCALS"));
                        for (index, decl) in body.clone().unwrap().local_decls() {
                            vector.push(format!("{index} = {}", decl.ty));
                        }
                        vector.push("".to_string());
                        local_node.set_label(vector.join("\\l").to_string().as_str());
                        local_node.set_style(Style::Filled);
                        local_node.set("color", "palegreen3");
```
image

jberthold (approved):

PR #78, opened 2025-05-04
Use git rev-parse to get master version

ehildenb:
Blocked on: https://github.com/runtimeverification/devops/pull/234
Blocked on: https://github.com/runtimeverification/mir-semantics/pull/561

This uses `git rev-parse` instead of relying on github action variables to get the version to push, as the prior attempt didn't end up including the label.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -23,7 +23,7 @@ jobs:
       - name: 'Update dependents'
         run: |
           set -x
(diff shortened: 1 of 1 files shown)


ehildenb:
We can see that the `curl` command didn't actually get the version in the prior attempt here: https://github.com/runtimeverification/stable-mir-json/actions/runs/14822311890/job/41610934736

image

dkcumming (approved):

PR #80, opened 2025-05-05
`MachineInfo` in JSON

dkcumming:
Added [MachineInfo](https://github.com/rust-lang/rust/blob/a2545fd6fc66b4323f555223a860c451885d1d2b/compiler/stable_mir/src/target.rs#L7-L12) to the JSON. This should give us the ability to use `usize` properly


jberthold (approved):
Let's add it for now but we should maybe revise later. I would have thought this should be placed near the top (next to name and  crate_id) but it might just be me.

gtrepta (approved):

PR #82, opened 2025-05-15
Add a list of field types to struct type metadata

### LINKED ISSUE #81: Add field type information to struct type metadata
Symbolic execution with struct types requires the struct fields to be generated (abstracted as variables) .
Without field types, the entire field list needs to be an opaque variable and leads to spurious branching.

Field types probably have to be retrieved by going via `adt_def` into the (single) variant metadata.

jberthold:
This adds the field types of the single variant of a `struct` `RigidTy::Adt` to the type metadata, similar to the `types` field of `TupleType`

Closes #81

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -987,6 +987,7 @@ pub enum TypeMetadata {
     StructType {
         name: String,
         adt_def: AdtDef,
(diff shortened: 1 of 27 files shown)


dkcumming (approved):
Looks great

PR #84, opened 2025-05-21
`TyKind` in `GlobalAlloc`s

dkcumming:
`GlobalAlloc`s need to be decoded to their types, this can be done easier by preprocessing the python (maybe in the future in the repo). But for the python to preprocess it will go easier with the TyKind available in the JSON.


ehildenb (approved):
Looks good overall, but you should probably wait to merge this one until you have a branch ready on KMIR that uses this and can confirm it's populating the `<globals>` cell correctly.

jberthold (approved):
LGTM

PR #85, opened 2025-05-23
Use mangled name for intrinsic symbols

ehildenb:


src/printer.rs
```diff
@@ -412,8 +412,8 @@ fn fn_inst_sym<'tcx>(
             let internal_inst = rustc_internal::internal(tcx, inst);
             let sym_type = if inst.is_empty_shim() {
                 NoOpSym(String::from(""))
-            } else if let Some(intrinsic_name) = inst.intrinsic_name() {
-                IntrinsicSym(intrinsic_name)
+            } else if let Some(_intrinsic_name) = inst.intrinsic_name() {
+                IntrinsicSym(inst.mangled_name())
```
ehildenb, line 416: Here we are trying to dump the mangled name of the intrinsics instead of just the short name.

tests/integration/programs/checked_arithmetic.rs
```diff
@@ -0,0 +1,21 @@
...
+fn checked_add_i32(a: i32, b: i32) {
+
+    precondition!(
+        a.checked_add(b).is_some(),
+         unsafe {
+            let result = a.unchecked_add(b);
+            assert!(result as i64 == a as i64 + b as i64)
+        }
+    );
+}
```
ehildenb, line 21: Example we are trying.

tests/integration/programs/checked_arithmetic.smir.json.expected
```diff
@@ -0,0 +1,3705 @@
...
+            "span": 45,
+            "spread_arg": null,
+            "var_debug_info": []
+          },
+          "id": 3,
+          "name": "std::intrinsics::cold_path"
+        }
+      },
+      "symbol_name": "_ZN4core10intrinsics9cold_path17h"
+    },
```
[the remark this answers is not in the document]
cds-amal: Hi @ehildenb , #138 is adding rustc_demangle for graph labels; if we also emit demangled names in the JSON output (as a lookup key alongside the mangled symbol), would that resolve the join problem you described?


Edit: got the wrong PR

PR #91, opened 2025-07-21
90 upgrade rustc dependency

### LINKED ISSUE #90: Upgrade compiler dependency to a recent nightly
The `rustc` dependency of `stable-mir-json` is currently to `nightly-2024-11-29`. 
We should upgrade it to be ahead of Anza's `rustc` fork and to not lose connection to latest `rustc` development.

However, we cannot upgrade to versions after early April due to a [temporary refactoring tweak that moved `stable_mir` into `rustc_smir`](https://github.com/rust-lang/rust/pull/139319). The import structure from this change onwards is still in place but is a temporary measure, so we should not upgrade to it (unless it remains in place for much longer).

Upgrading to the latest  nightly _before_ this change (or a reasonably close one) is a promising intermediate step.

UPDATE: https://github.com/rust-lang/rust/pull/143848 appears to have moved back what was formerly `stable_mir` into its place, but under the new name `rustc_public`.

jberthold:
The time seems right to make an upgrade of the rustc dependency:
* [`stable_mir` and `rustc_smir` just got renamed](https://github.com/rust-lang/rust/commit/6598c61725abb0e8179b8a5fab7a0ed1f7fcaf5f)
* quite a few small changes have accumulated in the compiler

Fixes #90


jberthold:
Current status: UI tests failures to investigate
  - 70x failing on `is_fn_ptr` assertion in `collect_alloc`, case `GlobalAlloc::Function`. Related to `FnOnce::call_once`
  - 22x failing on new unimplemented case `GlobalAlloc::TypeId` in `collect_alloc`

PR #94, opened 2025-07-24
92 extract type layouts

### LINKED ISSUE #92: Include essential layout information in type metadata
The type metadata emitted by stable-mir-json should include information about the type's layout in memory, for the purpose of decoding constants in the MIR (allocated statically or constant immediate values as operands).

In this context of decoding, we also need the fields of each variant of an `enum` type.

Currently, decoding is implemented for integral number types and bool in K code, but a general (recursive) method needs to be added.

jberthold:
Includes layout information for all type metadata (except primitive and function types).

The original Stable MIR data structures are dumped, probably containing some redundancy.

Fixes #92


dkcumming (approved):
Looks great!

PR #96, opened 2025-08-04
Add nix derivation

juliankuners:
This adds a nix derivation for stable-mir-json.

In addition, the unit and integration tests can be executed via nix by running `nix flake check`.

Also a nix formatter was added. It can be installed via `nix profile install nixpkgs#nixfmt` and run via `make format` or `nixfmt **/*.nix`.

A test workflow was added that builds and tests the nix derivation for both x86_64-linux and arm64-MacOS.

Finally, the derivation and its dependencies are published on the usual k-framework cachix nix binary caches.


juliankuners:
Once this was merged, stable-mir-json could be added to kup.

juliankuners:
@jberthold @dkcumming 
The nix MacOS build builds and work so far. But the integration tests generate slightly different output. It looks like the order of some elements is determinstically different on arm64 MacOS compared to x86_64 Linux.

Would you mind taking a look at this? I appended the diffs, from the remote MacOS machine that I have access to, to this comment.
[integration-tests-output.patch.txt](https://github.com/user-attachments/files/21586322/integration-tests-output.patch.txt)

jberthold:
> Would you mind taking a look at this? I appended the diffs, from the remote MacOS machine that I have access to, to this comment. Alternatively, we could simply not run the integration tests on the nix build, though I'd like to run some kind of smoke test at the very least.
> 
> [integration-tests-output.patch.txt](https://github.com/user-attachments/files/21586322/integration-tests-output.patch.txt)

Took a quick look... it seems like the item _order_ is different on MacOS. The jq normalisation script is currently not sorting `.functions` and `.items`, maybe we can just add a `sort` here? (or rather `.functions | sort_by(.[0])` and `.items | sort_by(.symbol_name)`? That might create some larger differences in the expected output but it should be OK. (I added better sorting to the `types` recently when I added the `layout`)

juliankuners:
Added `rv-nix-tools` as a dependency to specify the revision of `nixpkgs` that is used. `rv-nix-tools` is automatically updated by devops, see runtimeverification/devops#237.

juliankuners:
> > Would you mind taking a look at this? I appended the diffs, from the remote MacOS machine that I have access to, to this comment. Alternatively, we could simply not run the integration tests on the nix build, though I'd like to run some kind of smoke test at the very least.
> > [integration-tests-output.patch.txt](https://github.com/user-attachments/files/21586322/integration-tests-output.patch.txt)
> 
> Took a quick look... it seems like the item _order_ is different on MacOS. The [jq normalisation script](https://github.com/runtimeverification/stable-mir-json/blob/master/tests/integration/normalise-filter.jq#L13-L14) is currently not sorting `.functions` and `.items`, maybe we can just add a `sort` here? (or rather `.functions | sort_by(.[0])` and `.items | sort_by(.symbol_name)`? That might create some larger differences in the expected output but it should be OK. (I added better sorting to the `types` recently when I added the `layout`)

Thanks, this solves my issue. I used `sort` both times, because at the very least with `items` there are duplicate `symbol_name` entries, which results in non-determinism again, as tested.

PR #97, opened 2025-08-05
Add macOS support to stable-mir-json

Stevengre:
- Add operating system detection in record_ld_library_path function
- Use DYLD_LIBRARY_PATH on macOS instead of LD_LIBRARY_PATH
- Include Rust toolchain path in default library path for macOS
- Add DYLD_LIBRARY_PATH export in generated shell scripts for macOS
- Maintain backward compatibility with Linux and other systems


Stevengre:
Thank you for letting me know!

dkcumming (approved):
I think it looks good to go, what do you think @jberthold ?

PR #100, opened 2025-09-09
HOTFIX emit the correct Alloc ty for each alloc

jberthold:
The previous PR added a `ty` field in the `.allocs` records but it contained types that were _references to_ the actual type of the alloc in question, because of how they were collected.

This change fixes that problem for `Memory` allocations. Support for other allocations is not ready as of now, so fixing it only for `Memory`.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -658,7 +658,7 @@ fn get_prov_ty(pointer_ty: stable_mir::ty::Ty) -> Option<stable_mir::ty::Ty> {
 
 fn collect_alloc(
     val_collector: &mut InternedValueCollector,
-    ty: stable_mir::ty::Ty,
+    ty: stable_mir::ty::Ty, // NB this is the type of a _reference_ to the constant
     val: stable_mir::mir::alloc::AllocId,
 ) {
     let entry = val_collector.visited_allocs.entry(val);
@@ -677,7 +677,7 @@ fn collect_alloc(
                 pointed_ty.map(|ty| ty.kind()),
                 global_alloc
             );
-            entry.or_insert((ty, global_alloc.clone()));
+            entry.or_insert((pointed_ty.unwrap(), global_alloc.clone()));
             alloc.provenance.ptrs.iter().for_each(|(_, prov)| {
                 collect_alloc(val_collector, pointed_ty.unwrap(), prov.0);
             });


jberthold:
The fix here is incomplete, we have to change the `get_prov_type` calls and consider the _offset_ field in each provenance map entry. Example programs:
* [Small program from rustc UI tests that crashes: issue-15689-1.rs.txt](https://github.com/user-attachments/files/22249252/issue-15689-1.rs.txt)
* [Another program, creates two prov.map entries: enum-two-refs.rs.txt](https://github.com/user-attachments/files/22249251/enum-two-refs.rs.txt)
Both program are crashing the `stable-mir-json` extraction with `None.unwrap()` when  following a provenance map entry from a first allocation.
On master, the extraction succeeds but the wrong types are associated (references to the actual data type instead of the type itself, and wrong type information for the stored references in the enum's field).
Attaching the json here for reference: [issue-15689-ty-field.json](https://github.com/user-attachments/files/22249313/issue-15689-ty-field.json) [enum-two-refs.smir.json](https://github.com/user-attachments/files/22249310/enum-two-refs.smir.json)
To view the allocs and types, these `jq` commands are handy:

```
$ jq -c '.allocs[]' enum-two-refs.smir.json                 # see allocs
$ jq -c '.types[] | del(.. | .layout?)' enum-two-refs.smir.json # view type table without layout information
$ jq '.types[] | select(.[0] == 25)' enum-two-refs.smir.json    # view type 25 from the types map (in full + laid out)
```

tothtamas28:
So, in the `enum-two-refs.rs` example:

```rust
#[derive(PartialEq, Debug)]
enum Thing<'a> {
    Thing( &'a i16, &'a u16)
}

// ...

fn main() {
    assert_eq!(Thing::Thing(&1, &2), /* ... */);
    // ...
}
```

where the relevant `allocs` are

```json
{"alloc_id":0,"ty":25,"global_alloc":{"Memory":{"bytes":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"provenance":{"ptrs":[[0,8],[8,9]]},"align":8,"mutability":"Not"}}}
{"alloc_id":8,"ty":121,"global_alloc":{"Memory":{"bytes":[1,0],"provenance":{"ptrs":[]},"align":2,"mutability":"Not"}}}
{"alloc_id":9,"ty":121,"global_alloc":{"Memory":{"bytes":[2,0],"provenance":{"ptrs":[]},"align":2,"mutability":"Not"}}}
```

instead of

```json
{"alloc_id":0,"ty":"&Things","global_alloc":"..."}
{"alloc_id":8,"ty":"Things","global_alloc":"..."}
{"alloc_id":9,"ty":"Things","global_alloc":"..."}
```

We should have

```json
{"alloc_id":0,"ty":"Things","global_alloc":"..."}
{"alloc_id":8,"ty":"i16","global_alloc":"..."}
{"alloc_id":9,"ty":"u16","global_alloc":"..."}
```

(And of course, similarly for the other three data structure references.)

Is this the expected output?

PR #102, opened 2025-10-18
feat(printer.rs): support GlobalAlloc::Function with non-fn-ptr type

Stevengre:
We observed a function allocation reachable through a value whose type is not an explicit `fn ptr`. This can legitimately happen when code casts/transmutes function pointers, erases types via raw pointers, or otherwise flows a function address through a non-`fn ptr` type.

In such cases, try to recover the provenance-based pointed type at this (ty, offset) site and record that; if recovery fails, fall back to an opaque placeholder to keep traversal progressing.

### DIFF
--- a/tests/ui/failing.tsv
+++ b/tests/ui/failing.tsv
@@ -21,8 +21,8 @@ tests/ui/issues/issue-26641.rs	101
 tests/ui/label/label_break_value_desugared_break.rs	101
 tests/ui/macros/macro-meta-items.rs	101
 tests/ui/rfcs/rfc-2091-track-caller/caller-location-fnptr-rt-ctfe-equiv.rs	101
-tests/ui/rfcs/rfc-2091-track-caller/tracked-fn-ptr.rs	101
 tests/ui/rfcs/rfc-2091-track-caller/tracked-fn-ptr-with-arg.rs	101
+tests/ui/rfcs/rfc-2091-track-caller/tracked-fn-ptr.rs	101
 tests/ui/test-attrs/test-panic-while-printing.rs	101
 tests/ui/traits/const-traits/const-drop.rs	101
 tests/ui/try-block/issue-45124.rs	101


dkcumming:
Can you re-run the ui test suite in `tests/ui` I want to see if this changes how many tests we successfully serialise.

tests/ui/passing.tsv
```diff
@@ -1,21 +1,21 @@
...
-tests/ui/abi/extern/extern-pass-empty.rs
 tests/ui/abi/extern/extern-pass-FiveU16s.rs
 tests/ui/abi/extern/extern-pass-TwoU16s.rs
 tests/ui/abi/extern/extern-pass-TwoU32s.rs
 tests/ui/abi/extern/extern-pass-TwoU64s.rs
 tests/ui/abi/extern/extern-pass-TwoU8s.rs
+tests/ui/abi/extern/extern-pass-char.rs
+tests/ui/abi/extern/extern-pass-double.rs
+tests/ui/abi/extern/extern-pass-empty.rs
 tests/ui/abi/extern/extern-pass-u32.rs
```
jberthold: Uh oh, this must have been running on a MacOS, right? (I found some interesting [reading material about this problem](https://blog.zhimingwang.org/macos-lc_collate-hunt), perfect for distraction...).

https://github.com/runtimeverification/stable-mir-json/blob/a27a2b60914942ed95a62b742aa881ab2f332bf8/tests/ui/remake_ui_tests.sh#L77-L79
We should put `LC_ALL=C` into the `remake_ui_tests.sh` script (at the sorting step) so this gets sorted the same way on all platforms to avoid diff noise.

Stevengre: @jberthold  My order doesn't change with the added env. I'll merge this first and investigate this order problem later.
- Here is the corresponding issue: https://github.com/runtimeverification/stable-mir-json/issues/103

Issue #83, opened 2025-05-20
Graph Improvement: GlobalAllocs


dkcumming:
It would be good to display the mapping of `GlobalAlloc`s in the graph, and perhaps noting which constants are promoted. And noting when a place is referencing a `GlobalAlloc` in a `Statement` or `Terminator`

PR #111, opened 2025-11-30
Pr/83

cds-amal:
Summary

This PR enhances the MIR graph visualization with richer allocation information and adds D2 as an alternative output format.

Key changes:

- Index-first architecture: Build `AllocIndex`, `TypeIndex`, and `GraphContext` upfront before graph traversal, enabling context-aware rendering throughout
- **ALLOCS legend**: New yellow info node showing all `GlobalAlloc` entries with decoded values (strings as ASCII, small integers as numeric values)
- Enhanced constant labels: Constants now show provenance references like `const [alloc0: Int(I32) = 42]` instead of opaque `const ?_i32`
- D2 output format (--d2 flag): Alternative to DOT with modern diagramming, viewable in browser at https://play.d2lang.com

## Addresses
#83 

## Example

Before:
```rust
const ?_i32
```

After:
```rust
const [alloc0: Int(I32) = 42]
```

## New CLI usage

```console
cargo run -- --dot file.rs   # Graphviz DOT (existing)
cargo run -- --d2 file.rs    # D2 format (new)
```

## New Make targets
```console
make dot    # Generate all dot files
make svg    # Generate all svg files
make png    # Generate all png files
make d2      # Generate all D2 files
```

## Test plan

- [ ] Run --dot on test programs, verify ALLOCS legend appears
- [ ] Run --d2 on test programs, verify output renders in D2 playground
- [ ] Verify make integration-test passes

### DIFF
--- a/src/mk_graph/index.rs
+++ b/src/mk_graph/index.rs
@@ -140,7 +140,7 @@ impl AllocEntry {
         let (kind, description) = match info.global_alloc() {
             GlobalAlloc::Memory(alloc) => {
                 let bytes = &alloc.bytes;
-                let is_str = ty_name.contains("str") || ty_name.contains("&str");
+                let is_str = ty_name.contains("str");
 
                 // Convert Option<u8> bytes to actual bytes for display
                 let concrete_bytes: Vec<u8> = bytes.iter().filter_map(|&b| b).collect();


src/mk_graph/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    match tcx.output_filenames(()).path(OutputType::Mir) {
+        OutFileName::Stdout => {
+            write!(io::stdout(), "{}", smir_d2).expect("Failed to write smir.d2");
+        }
+        OutFileName::Real(path) => {
+            let out_path = path.with_extension("smir.d2");
+            let mut b = io::BufWriter::new(
+                File::create(&out_path)
+                    .unwrap_or_else(|e| panic!("Failed to create {}: {}", out_path.display(), e)),
```
dkcumming:
```suggestion
                    .unwrap_or_else(|e| panic!("Failed to create {}:\n\n{}", out_path.display(), e)),
```

Makefile
```diff
@@ -69,3 +69,47 @@ test-ui:
...
+.PHONY: dot svg png d2 clean-graphs
+
+OUTDIR_DOT=output-dot
+OUTDIR_SVG=output-svg
+OUTDIR_PNG=output-png
+OUTDIR_D2=output-d2
+
+clean-graphs:
+	@rm -rf $(OUTDIR_DOT) $(OUTDIR_SVG) $(OUTDIR_PNG) $(OUTDIR_D2)
+
```
dkcumming:
```suggestion
clean-graphs:
	@rm -rf $(OUTDIR_DOT) $(OUTDIR_SVG) $(OUTDIR_PNG) $(OUTDIR_D2)

check-graphviz:
	@command -v dot >/dev/null 2>&1 || { echo "Error: graphviz not installed. Install with: apt install graphviz"; exit 1; }

```

src/mk_graph/index.rs
```diff
@@ -0,0 +1,219 @@
...
+                let concrete_bytes: Vec<u8> = bytes.iter().filter_map(|&b| b).collect();
+
+                let desc = if is_str && concrete_bytes.iter().all(|b| b.is_ascii()) {
+                    let s: String = concrete_bytes
+                        .iter()
+                        .take(20)
+                        .map(|&b| b as char)
+                        .collect::<String>()
+                        .escape_default()
+                        .to_string();
```
dkcumming: I think this could be changed to work for utf8. Although the code would need to change from `take(20)` because you are not guaranteed that a `char` will fit 1 byte so 20 might be in the middle of a char. I don't know if this is needed for this PR though, I would be happy to make an issue for it to handle utf8

dkcumming (commented):
Awesome stuff, I'm loving the whole thing. Some minor fix ups, and I would personally like to add the fast fail if `dot` is not installed with an informative message.

Do you think you could rebase and consider the suggestions?

Also what do you think about the ut8? I think we leave it for another issue right now so we get this in faster (sorry for my delay in getting this reviewed)

src/mk_graph/index.rs
```diff
@@ -0,0 +1,219 @@
...
+impl AllocEntry {
+    pub fn from_alloc_info(info: &AllocInfo, type_index: &TypeIndex) -> Self {
+        let alloc_id = info.alloc_id().to_index() as u64;
+        let ty = info.ty();
+        let ty_name = type_index.get_name(ty);
+
+        let (kind, description) = match info.global_alloc() {
+            GlobalAlloc::Memory(alloc) => {
+                let bytes = &alloc.bytes;
+                let is_str = ty_name.contains("str") || ty_name.contains("&str");
```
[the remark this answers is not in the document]
cds-amal: oh yea! makes sense.

src/mk_graph/index.rs
cds-amal: One does not casually reason about the wide and skinny bytes in a PR; when implementing this, I wanted to avoid the issue and should have called it out. Would appreciate if we could raise an issue, so a fix is more focused, that would be a good-first-contrib category.

Issue #55, opened 2025-03-13
Add functions passed as (function pointer) arguments to the `functions` lookup array


jberthold:
Related: https://github.com/runtimeverification/mir-semantics/issues/488

Sometimes functions are called via function pointers instead of as constants in MIR. Therefore, the `functions` array must be populated with these function items that are (potentially) called via pointers (currently only constants from `func` in `TerminatorKind::Call` are added to the array).

Rather than trying to go backwards from the `Call` terminator, this should be done by inspecting arguments of function calls to determine whether they are function pointers, and adding those (they may _potentially_ be called later after being passed on to other callees in a chain of calls).

PR #112, opened 2025-12-15
fix(printer.rs): include ZeroSized FnDef consts in functions

Stevengre:
- Problem: stable-mir-json misses function items (FnDef) that appear as ConstantKind::ZeroSized when passed as values (e.g. Option::map(u64::from_le_bytes)), so the emitted functions list is incomplete.
- Fix: In InternedValueCollector::visit_mir_const, detect ZeroSized constants whose type is FnDef, resolve an Instance (prefer resolve_for_fn_ptr, fallback to resolve), and add it to the link map with ItemSource(FPTR) so it is emitted in functions.
- Robustness: Avoid panicking on inconsistent symbol mappings by skipping insertion when the existing LinkMapKey already maps to a different symbol (direct-call vs reify shim cases).
- Related: #55 (and downstream https://github.com/runtimeverification/mir-semantics/issues/891 https://github.com/runtimeverification/mir-semantics/issues/488).


jberthold (approved):
LGTM, thanks for adjusting!

PR #119, opened 2026-02-18
Filter `std::rt::lang_start` from graphs

dkcumming:
This PR allows us to set a `SKIP_LANG_START=1` env var to remove the `std::rt::lang_start` items from the graph output. Perhaps sometimes they are useful, but I have not ever needed to look at them and would prefer them to be removed from the graph.

I had a bit of trouble getting the correct items for the BFS and I used claude to help for full disclosure, however I have vetted the code and agree with it. Also the output has been tested and it does remove the items correctly. I will post before and after of a contrived program that contains `black_box` to show that the algorithm will only remove items exclusively downstream of `std::rt::lang_start`.

```rust
fn main() {
    let x = std::hint::black_box(42);
    std::hint::black_box(x);
}
```

### Before
image

### After
image

### DIFF
--- a/src/mk_graph/output/d2.rs
+++ b/src/mk_graph/output/d2.rs
@@ -1,20 +1,32 @@
 //! D2 diagram format output for MIR graphs.
 
+use std::collections::HashSet;
+
 use crate::compat::stable_mir;
 use stable_mir::mir::TerminatorKind;
 
-use crate::printer::SmirJson;
-use crate::MonoItemKind;
+use crate::printer::{MonoItemKind, SmirJson};
 
 use crate::mk_graph::context::GraphContext;
 use crate::mk_graph::util::{
     escape_d2, is_unqualified, name_lines, short_name, terminator_targets,
 };
+use crate::mk_graph::{compute_lang_start_exclusions, skip_lang_start};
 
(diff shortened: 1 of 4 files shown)


src/mk_graph/output/d2.rs
```diff
@@ -124,7 +139,7 @@ fn render_d2_call_edges(
         let Some(callee_name) = ctx.resolve_call_target(func) else {
             continue;
         };
-        if !is_unqualified(&callee_name) {
+        if !is_unqualified(&callee_name) || excluded.contains(&callee_name) {
             continue;
         }
```
cds-amal: Want to point out that we have to thread `excluded` into this function to filter a target. Seems. like a smell -- Not asking for a change :)

cds-amal (commented):
This is really nice @dkcumming, I love the story in your commits. I feel strongly about including a test to surface regressions, as I expect there are many more features to come.

There are no integration or unit tests for the `SKIP_LANG_START` behavior. The existing `.smir.json.expected` files all contain lang_start items, so there's good data to work from. At minimum, an integration test that runs with `SKIP_LANG_START=1` and verifies the output lacks lang_start entries would give confidence this doesn't regress.

PR #129, opened 2026-02-28
fix(printer): record closure Ty in type table after instance traversal

Stevengre:
### Summary
- In `TyCollector::visit_ty` (`src/printer.rs`), after successful closure instance traversal, record closure `Ty` metadata in `types`.
- Update integration golden files (`tests/integration/programs/*.smir.json.expected`) to match emitted closure `FunType` metadata.
- Keep `Makefile` integration path as `TESTDIR=tests/integration/programs` to keep closure fixture output paths stable.

### Context
In the closure arm of `TyCollector::visit_ty`, the code resolved and visited the closure instance but returned without inserting the closure `Ty` into `types`. Other type families were inserted; closures were the exception.

This made downstream closure type lookup incomplete (notably for kmir proof paths that consume exported type metadata). Recording closure `Ty` closes that gap.

Related split: static/vtable alloc fallback changes were moved out to [stable-mir-json#131](https://github.com/runtimeverification/stable-mir-json/pull/131) so this PR stays focused on closure metadata.

Dependency context: this PR is depended on by [mir-semantics#957](https://github.com/runtimeverification/mir-semantics/pull/957).


dkcumming (approved):
LGTM

PR #132, opened 2026-03-02
Refactor MIR graph rendering to use a unified traversal via GraphBuilder

0xh4ty:
### Summary

This PR introduces a format-agnostic MIR graph traversal based on a new `GraphBuilder` abstraction. Traversal semantics are centralized in a single implementation, while individual renderers are responsible only for presentation.

As an initial consumer, the D2 renderer has been ported to this new traversal behind a parallel code path. The legacy D2 implementation is retained for now to allow safe comparison and validation.

---

### What changed

* Introduced a `GraphBuilder` trait describing semantic graph events
* Added a single generic traversal (`render_graph`) that owns MIR graph structure and traversal order
* Refactored the D2 renderer to implement `GraphBuilder`
* Moved all semantic stringification to `GraphContext`
* Traversal now passes full MIR structures (statements, terminators, call arguments) to renderers
* Wired the new D2 path in parallel to the legacy one for validation
* Verified that legacy and new D2 outputs are byte-for-byte identical across the test suite
* Code passes both `cargo fmt` and `cargo clippy`

---

### Why this change

Previously, each renderer duplicated traversal logic, which made the code harder to reason about and maintain. Centralizing traversal ensures:

* a single source of truth for graph semantics
* easier maintenance and future refactors
* cleaner renderer implementations
* safer extension to additional output formats

---

### Validation and testing

The new traversal can be exercised by enabling the experimental D2 path:

```bash
SMIR_D2_NEW=1 cargo run -- --d2 tests/integration/programs/fibonacci.rs
```

To compare outputs against the legacy renderer:

```bash
git diff --no-index output-d2/fibonacci.smir.d2 fibonacci.smir.d2
```

No differences were observed across the existing integration test suite.

---

### Status and follow-ups

* The legacy D2 renderer is intentionally kept for now
* Once this approach is validated, the D2 renderer can be switched to use the new traversal by default in follow-up commits.
* Refactoring the DOT renderer to the same model will follow in a separate PR

---

### Notes for reviewers

This PR is a structural refactor only. Traversal semantics and output behavior are unchanged, and the parallel wiring is intended to make review and validation straightforward.

---

> **Note on function IDs**
> The legacy D2 renderer uses `short_name`-based function IDs, which are prone to collisions in cases such as multiple monomorphizations of the same generic.
>
> This behavior is intentionally preserved for now so that the new traversal can be validated via byte-for-byte output comparison. Collision avoidance will be addressed in follow-up commits by incorporating body hashes into function IDs.

### DIFF
--- a/src/mk_graph/output/d2.rs
+++ b/src/mk_graph/output/d2.rs
@@ -1,7 +1,7 @@
 //! D2 diagram format output for MIR graphs.
 
 use crate::compat::stable_mir;
-use stable_mir::mir::{Operand, Statement, Terminator, TerminatorKind};
+use stable_mir::mir::{TerminatorKind};
 
 use crate::printer::SmirJson;
 use crate::MonoItemKind;
@@ -12,27 +12,23 @@ use crate::mk_graph::util::{
 };
 
 use crate::mk_graph::traverse::render_graph;
-use crate::mk_graph::traverse::GraphBuilder;
+use crate::mk_graph::traverse::{GraphBuilder, RenderedFunction};
 
 // =============================================================================
 // D2 Builder
 // =============================================================================
 
-pub struct D2Builder<'a> {
-    ctx: &'a GraphContext,
+pub struct D2Builder {
     buf: String,
 }
 
-impl<'a> D2Builder<'a> {
-    pub fn new(ctx: &'a GraphContext) -> Self {
-        Self {
-            ctx,
-            buf: String::new(),
-        }
+impl D2Builder {
+    pub fn new() -> Self {
+        Self { buf: String::new() }
     }
 }
 
-impl GraphBuilder for D2Builder<'_> {
+impl GraphBuilder for D2Builder {
     type Output = String;
 
     fn begin_graph(&mut self, _name: &str) {
@@ -40,65 +36,83 @@ impl GraphBuilder for D2Builder<'_> {
     }
 
     fn alloc_legend(&mut self, lines: &[String]) {
+
         self.buf.push_str("ALLOCS: {\n");
         self.buf.push_str("  style.fill: \"#ffffcc\"\n");
         self.buf.push_str("  style.stroke: \"#999999\"\n");
-        let legend_text = lines
+
+        let text = lines
             .iter()
-            .map(|s| escape_d2(s))
+            .map(|l| escape_d2(l))
             .collect::<Vec<_>>()
             .join("\\n");
-        self.buf
-            .push_str(&format!("  label: \"{}\"\n", legend_text));
+
+        self.buf.push_str(&format!("  label: \"{}\"\n", text));
         self.buf.push_str("}\n\n");
     }
 
-    fn type_legend(&mut self, _: &[String]) {}
+    fn type_legend(&mut self, _lines: &[String]) {}
 
-    fn begin_function(&mut self, id: &str, label: &str, _is_local: bool) {
-        self.buf.push_str(&format!("{}: {{\n", id));
+    fn external_function(&mut self, id: &str, name: &str) {
+
         self.buf
-            .push_str(&format!("  label: \"{}\"\n", escape_d2(label)));
+            .push_str(&format!("{}: \"{}\"\n", id, escape_d2(name)));
+    }
+
+    fn render_function(&mut self, func: &RenderedFunction) {
+
+        self.buf.push_str(&format!("{}: {{\n", func.id));
+        self.buf
+            .push_str(&format!("  label: \"{}\"\n", escape_d2(&func.display_name)));
         self.buf.push_str("  style.fill: \"#e0e0ff\"\n");
-    }
 
-    fn block(&mut self, _fn_id: &str, idx: usize, stmts: &[Statement], terminator: &Terminator) {
-        let mut label = format!("bb{}:", idx);
-        for stmt in stmts {
-            let s = self.ctx.render_stmt(stmt);
-            label.push_str(&format!("\\n{}", escape_d2(&s)));
+        for block in &func.blocks {
+
+            let mut label = format!("bb{}:", block.idx);
+
+            for stmt in &block.stmts {
+                label.push_str(&format!("\\n{}", escape_d2(stmt)));
+            }
+
+            label.push_str(&format!("\\n---\\n{}", escape_d2(&block.terminator)));
+
+            self.buf
+                .push_str(&format!("  bb{}: \"{}\"\n", block.idx, label));
         }
-        let term_str = self.ctx.render_terminator(terminator);
-        label.push_str(&format!("\\n---\\n{}", escape_d2(&term_str)));
 
-        self.buf.push_str(&format!("  bb{}: \"{}\"\n", idx, label));
-    }
+        for block in &func.blocks {
+            for (target, _) in &block.cfg_edges {
+                self.buf
+                    .push_str(&format!("  bb{} -> bb{}\n", block.idx, target));
+            }
+        }
 
-    fn block_edge(&mut self, _fn_id: &str, from: usize, to: usize, _label: Option<&str>) {
-        self.buf.push_str(&format!("  bb{} -> bb{}\n", from, to));
-    }
-
-    fn call_edge(
-        &mut self,
-        fn_id: &str,
-        block: usize,
-        callee_id: &str,
-        callee_name: &str,
-        _args: &[Operand],
-    ) {
-        self.buf
-            .push_str(&format!("{}: \"{}\"\n", callee_id, escape_d2(callee_name)));
-        self.buf
-            .push_str(&format!("{}.style.fill: \"#ffe0e0\"\n", callee_id));
-        self.buf
-            .push_str(&format!("{}.bb{} -> {}: call\n", fn_id, block, callee_id));
-    }
-
-    fn end_function(&mut self, _id: &str) {
         self.buf.push_str("}\n\n");
+
+        for edge in &func.call_edges {
+
+            self.buf.push_str(&format!(
+                "{}: \"{}\"\n",
+                edge.callee_id,
+                escape_d2(&edge.callee_name)
+            ));
+
+            self.buf.push_str(&format!(
+                "{}.style.fill: \"#ffe0e0\"\n",
+                edge.callee_id
+            ));
+
+            self.buf.push_str(&format!(
+                "{}.bb{} -> {}: call\n",
+                func.id,
+                edge.block_idx,
+                edge.callee_id
+            ));
+        }
     }
 
     fn static_item(&mut self, id: &str, name: &str) {
+
         self.buf
             .push_str(&format!("{}: \"{}\" {{\n", id, escape_d2(name)));
         self.buf.push_str("  style.fill: \"#e0ffe0\"\n");
@@ -106,8 +120,10 @@ impl GraphBuilder for D2Builder<'_> {
     }
 
     fn asm_item(&mut self, id: &str, content: &str) {
-        let asm_text = escape_d2(&content.lines().collect::<String>());
-        self.buf.push_str(&format!("{}: \"{}\" {{\n", id, asm_text));
+
+        let text = escape_d2(&content.lines().collect::<String>());
+
+        self.buf.push_str(&format!("{}: \"{}\" {{\n", id, text));
         self.buf.push_str("  style.fill: \"#ffe0ff\"\n");
         self.buf.push_str("}\n\n");
     }
@@ -137,8 +153,7 @@ impl SmirJson<'_> {
 
     /// Convert the MIR to D2 using GraphBuilder traversal (experimental)
     pub fn to_d2_file_new(&self) -> String {
-        let ctx = GraphContext::from_smir(self);
-        render_graph(self, D2Builder::new(&ctx))
+        render_graph(self, D2Builder::new())
     }
 }
 
--- a/src/mk_graph/output/d2.rs
+++ b/src/mk_graph/output/d2.rs
@@ -11,7 +11,25 @@ use crate::mk_graph::util::{
     escape_d2, is_unqualified, name_lines, short_name, terminator_targets,
 };
 
-impl SmirJson {
+// =============================================================================
+// D2 Builder
+// =============================================================================
+
+struct D2Builder {
+    buf: String,
+}
+
+impl D2Builder {
+    fn new() -> Self {
+        Self { buf: String::new() }
+    }
+}
+
+// =============================================================================
+// Public entry point
+// =============================================================================
+
+impl SmirJson<'_> {
     /// Convert the MIR to D2 diagram format
     pub fn to_d2_file(self) -> String {
         let ctx = GraphContext::from_smir(&self);
--- a/src/mk_graph/output/d2.rs
+++ b/src/mk_graph/output/d2.rs
@@ -20,19 +20,7 @@ impl SmirJson {
         output.push_str("direction: right\n\n");
         render_d2_allocs_legend(&ctx, &mut output);
 
-        for item in self.items {
-            match item.mono_item_kind {
-                MonoItemKind::MonoItemFn { name, body, .. } => {
-                    render_d2_function(&name, body.as_ref(), &ctx, &mut output);
-                }
-                MonoItemKind::MonoItemGlobalAsm { asm } => {
-                    render_d2_asm(&asm, &mut output);
-                }
-                MonoItemKind::MonoItemStatic { name, .. } => {
-                    render_d2_static(&name, &mut output);
-                }
-            }
-        }
+        render_d2_items(&self.items, &ctx, &mut output);
 
         output
     }
@@ -41,6 +29,22 @@ impl SmirJson {
 // =============================================================================
 // D2 Rendering Helpers
 // =============================================================================
+
+fn render_d2_items(items: &[crate::printer::Item], ctx: &GraphContext, out: &mut String) {
+    for item in items {
+        match &item.mono_item_kind {
+            MonoItemKind::MonoItemFn { name, body, .. } => {
+                render_d2_function(name, body.as_ref(), ctx, out);
+            }
+            MonoItemKind::MonoItemGlobalAsm { asm } => {
+                render_d2_asm(asm, out);
+            }
+            MonoItemKind::MonoItemStatic { name, .. } => {
+                render_d2_static(name, out);
+            }
+        }
+    }
+}
 
 fn render_d2_allocs_legend(ctx: &GraphContext, out: &mut String) {
     let legend_lines = ctx.allocs_legend_lines();
--- a/src/mk_graph/output/d2.rs
+++ b/src/mk_graph/output/d2.rs
@@ -11,6 +11,8 @@ use crate::mk_graph::util::{
     escape_d2, is_unqualified, name_lines, short_name, terminator_targets,
 };
 
+use crate::mk_graph::traverse::GraphBuilder;
+
 // =============================================================================
 // D2 Builder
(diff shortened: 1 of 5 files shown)


Stevengre:
LGTM! Would you remove the previous implementation before merging? Or just in another PR?

@dkcumming what do you think?

0xh4ty:
Thanks! I’ll remove the legacy D2 implementation in this PR itself.

src/mk_graph/traverse.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    fn asm_item(&mut self, id: &str, content: &str);
+
+    fn finish(self) -> Self::Output;
+}
+
+/// Format-agnostic MIR graph traversal.
+/// Owns traversal order and graph semantics, delegates rendering to `GraphBuilder`.
+pub fn render_graph<B: GraphBuilder>(smir: &SmirJson, mut builder: B) -> B::Output {
+    let ctx = GraphContext::from_smir(smir);
```
Stevengre, line 90:
```rust
    pub fn to_d2_file_new(&self) -> String {
        let ctx = GraphContext::from_smir(self);
        render_graph(self, D2Builder::new(&ctx))
    }
```

It seems not efficient. It builds `GraphContext` twice.

src/mk_graph/traverse.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Traverses blocks, CFG edges, and call edges without renderer-specific logic.
+fn render_function<B: GraphBuilder>(
+    ctx: &GraphContext,
+    builder: &mut B,
+    name: &str,
+    body: Option<&stable_mir::mir::Body>,
+) {
+    let fn_id = short_name(name);
+    let label = name_lines(name);
+    let is_local = true;
```
Stevengre: Should be `body.is_some()` ?

src/mk_graph/traverse.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use crate::mk_graph::context::GraphContext;
+use crate::mk_graph::util::{is_unqualified, name_lines, short_name, terminator_targets};
+
+/// Format agnostic graph sink.
+/// Implemented by all renderers.
+pub trait GraphBuilder {
+    type Output;
+
+    fn begin_graph(&mut self, name: &str);
```
Stevengre, line 67: I'd like to have some description here.

src/mk_graph/traverse.rs
0xh4ty: Good catch. I’ll move `GraphContext` construction out of `render_graph` so it is built only once and passed through.

src/mk_graph/traverse.rs
0xh4ty: Yes, that should be `body.is_some()`. I’ll fix that.

src/mk_graph/traverse.rs
0xh4ty: Got it. I’ll expand the doc comment.

dkcumming (commented):
@0xh4ty this is great work! I love what you have done so far! I think it's time to go all in and get the old stuff out and hook up the full implementation of d2 and dot (okay if that is another PR if necessary). I think isolating the shared logic out in the traverse module is a great improvement. Do you think you are fine to do the full conversion? Also I think Jianhong had some great feedback too

src/mk_graph/traverse.rs
cds-amal: this `true` assignment is unconditional; is it needed?

src/mk_graph/output/d2.rs
```diff
@@ -11,6 +11,116 @@ use crate::mk_graph::util::{
...
+            let s = self.ctx.render_stmt(stmt);
+            label.push_str(&format!("\\n{}", escape_d2(&s)));
+        }
+        let term_str = self.ctx.render_terminator(terminator);
+        label.push_str(&format!("\\n---\\n{}", escape_d2(&term_str)));
+
+        self.buf.push_str(&format!("  bb{}: \"{}\"\n", idx, label));
+    }
+
+    fn block_edge(&mut self, _fn_id: &str, from: usize, to: usize, _label: Option<&str>) {
```
cds-amal: Same issue here: `block_edge` bakes D2's arrow syntax into the implementation, and the `Option<&str>` label parameter (which it looks like D2Builder ignores?) is a tell that the signature was shaped around what D2 needs today, with a speculative parameter tacked on for a future renderer.

This is the second method where format-specific concerns leak through the trait boundary, so it's worth stepping back and asking: what's the trait actually buying us?

A well-drawn trait boundary gives you two things: the code above it can change without touching the code below (new traversal logic doesn't require touching every renderer), and the code below can change without touching the code above (new output format, same traversal). But that only works if the boundary is at the right level of abstraction. When trait methods receive raw MIR types or carry format-specific assumptions, every new renderer has to understand the same internals, and the trait becomes ceremony rather than insulation.

The fix is the same as for block: have the traversal own the "what" (which blocks connect, what the edge represents) and pass structured, pre-rendered data. Let each renderer own the "how" (syntax, escaping, layout). That way the trait boundary actually earns its keep.

cds-amal (changes requested):
Good work @0xh4ty . @Stevengre  gave some excellent feedback, and I added a few notes on the trait boundary design. 

The traversal extraction is the right idea, and the hard part (generic traversal order, clean separation of items/blocks/edges) is already solid. The main thing to tighten up is where the format-agnostic boundary actually sits: a couple of the trait methods (`block`, `block_edge`) still carry D2-specific assumptions, which means the next renderer would inherit those assumptions rather than being free to do its own thing. 

The comments above walk through the specifics. Once those methods pass structured data instead of raw MIR types and format-specific conventions, this will be a clean foundation for DOT, Mermaid, and whatever else comes next.

I also want to see more comments, as I want this project to be usable via `cargo doc --open`, one more sign of a mature Cargo library.

Finally, my vote is to rip out the old, and use this as the way forward. Though, you will have to make sure the operational parts remain consistent so we can generate graphs :)

src/mk_graph/traverse.rs
0xh4ty: This will be removed.

src/mk_graph/output/d2.rs
0xh4ty: Thanks, that makes sense. The label parameter was speculative and is not used by the current renderers, so I will remove it to avoid leaking format assumptions into the trait. I will keep traversal responsible for edge semantics and pass structured, pre-rendered data to builders, while the renderers handle syntax and layout.

cds-amal:
Hey @0xh4ty ,  I had the opportunity to think this through some more and, well, here you go :)

First: the `GraphBuilder` trait is the right abstraction. Separating "walk the graph" from "emit format-specific syntax" is exactly where we want to end up, and the incremental rollout (old path kept, new path behind an env var) is a clean way to get there without breaking anything.

The question I want to think through is: what happens when we go to add DOT, Mermaid, Markdown, or other formats on top of this? Right now, each builder holds a `&GraphContext`, calls `ctx.render_stmt()` / `ctx.render_terminator()` / `ctx.render_operand()` itself, and carries a lifetime parameter. That means every new format re-implements the same rendering logic. And the trait's surface area will grow as formats discover they need additional hooks (`locals_table`, `empty_body`, `external_function`, etc.); each of those becomes a new method that every builder has to implement, even if it's a no-op.

The thing is, the current `GraphBuilder` is quietly conflating two jobs: (1) "what is the graph structure?" (nodes, edges, containment) and (2) "how do I turn a `Statement` or `Terminator` into a human-readable string?" Every builder currently does both. If we pull those apart, adding a new format becomes "just format pre-rendered strings into your syntax" rather than "re-implement MIR rendering *and* format-specific syntax."

So here's a concrete idea: instead of the driver pushing raw stable_mir types to the builder via a sequence of granular calls (`begin_function` / `block` / `block_edge` / `call_edge` / `end_function`), the driver produces a `RenderedFunction` struct that the builder receives as one cohesive unit:

```rust
/// A single basic block with pre-rendered content and structural edges.
struct RenderedBlock<'a> {
    idx: usize,
    stmts: Vec<String>,              // pre-rendered via GraphContext
    terminator: String,               // pre-rendered
    raw_terminator: &'a Terminator,   // escape hatch for structural inspection
    cfg_edges: Vec<(usize, Option<String>)>,  // (target_block, optional label)
}

/// A fully analyzed function, ready for format-specific rendering.
struct RenderedFunction<'a> {
    id: String,                       // stable ID (short_name + body hash)
    display_name: String,             // name_lines() output for labels
    is_local: bool,
    locals: Vec<(usize, String)>,     // (index, type_with_layout string)
    blocks: Vec<RenderedBlock<'a>>,
    call_edges: Vec<CallEdge>,        // resolved callee IDs and pre-rendered args
}

struct CallEdge {
    block_idx: usize,
    callee_id: String,
    callee_name: String,
    rendered_args: String,
}
```

The traversal driver does all the `GraphContext` work (rendering statements, resolving callees, computing function IDs) and hands `RenderedFunction` values to the builder. The trait itself simplifies:

```rust
trait GraphBuilder {
    type Output;

    fn begin_graph(&mut self, name: &str);
    fn alloc_legend(&mut self, lines: &[String]);
    fn type_legend(&mut self, lines: &[String]);
    fn external_function(&mut self, id: &str, name: &str);
    fn render_function(&mut self, func: &RenderedFunction);
    fn static_item(&mut self, id: &str, name: &str);
    fn asm_item(&mut self, id: &str, content: &str);
    fn finish(self) -> Self::Output;
}
```

What does this actually buy us? A few things:

No lifetime on builders. `RenderedFunction` owns its strings, so the builder doesn't need `&GraphContext` or a lifetime parameter. `D2Builder` becomes a plain struct, not `D2Builder<'a>`.

Rendering logic written once. The driver calls `ctx.render_stmt()`, `ctx.render_terminator()`, `ctx.render_operand()`. No builder ever touches `GraphContext`. You write the rendering code once; every format gets it for free.

Still flexible where it matters. The `raw_terminator` field (and potentially `raw_stmts` if we add it later) is an escape hatch for builders that need structural information; say, coloring blocks differently based on terminator kind, or visually distinguishing `Call` terminators from `Assert`. Builders that don't need this just ignore it.

Fewer trait methods. The granular push-based sequence (`begin_function` / `block` / `block_edge` / `end_function`) collapses into one `render_function(&RenderedFunction)` call. Builders that want the push style can still iterate `func.blocks` internally; the trait just doesn't force a particular traversal order.

New hooks emerge as data, not methods. Instead of adding `locals_table()`, `empty_body()`, `label_to_entry_edge()` as separate trait methods that every builder must implement (even as no-ops), these become fields on `RenderedFunction` that builders can inspect or ignore. `func.locals` is empty if there are no locals; `func.blocks` is empty if there's no body. The builder decides how to handle each case in its own `render_function` impl. No trait churn.

One more thing worth noting: function IDs. The current branch uses `short_name(name)` for graph node IDs, which works for simple cases but will collide when two monomorphizations of the same generic (say, `foo::<i32>` and `foo::<u64>`) produce the same short name. A `make_fn_id` helper that incorporates a body hash as a tiebreaker fixes this; with the `RenderedFunction` approach, that helper lives in the driver and produces `func.id`, so no format has to worry about it:

```rust
/// Generate a stable, unique ID from a symbol name and body.
/// Handles the case where multiple monomorphizations of the same
/// generic produce the same short_name by incorporating a body hash.
fn make_fn_id(name: &str, body: &Body) -> String { ... }
```

In terms of concrete next steps, I'd suggest:

1. Introduce `RenderedBlock`, `RenderedFunction`, and `CallEdge` structs (could live in `traverse.rs` alongside the driver).
2. Have `render_graph` produce `RenderedFunction` values and pass them to the builder.
3. Simplify `D2Builder`: drop the `&'a GraphContext` field and lifetime parameter, implement the narrower `render_function(&RenderedFunction)` method.
4. Verify the D2 output is identical (should be a pure refactor; no output change).
5. As a stretch goal: try implementing a second format (DOT or Mermaid) on top of the new trait to validate that it's genuinely easy to add.

Of course, feel free to push back, if you have a different roadmap in mind.

0xh4ty:
Thanks for the detailed explanation. That separation between graph structure and MIR string rendering makes sense. I will refactor the driver to produce `RenderedFunction`,  `RenderedBlock` and `CallEdge` data with pre-rendered strings and simplify the `GraphBuilder` trait to operate on that representation. This should also remove the need for `GraphContext` inside builders and eliminate the lifetime parameter. I’ll update the PR accordingly.

0xh4ty:
I pushed an update addressing the review feedback.

Main changes:

• Introduced `RenderedFunction`, `RenderedBlock`, and `CallEdge` as the traversal output structures.
• The traversal layer now performs MIR rendering using `GraphContext` and constructs these structures.
• `GraphBuilder` implementations now receive pre-rendered data and are responsible only for format-specific formatting.
• Builders no longer depend on `GraphContext`, removing the lifetime parameter and avoiding duplicated MIR rendering logic across renderers.
• Function IDs now incorporate a body hash to avoid collisions between monomorphizations.
• The legacy D2 implementation has been removed and the renderer now uses the new traversal by default.

The D2 output remains the same except for the expected function ID change caused by the body hash used for collision avoidance.

src/mk_graph/util.rs
```diff
@@ -280,3 +280,29 @@ pub fn terminator_targets(term: &Terminator) -> Vec<usize> {
         }
     }
 }
+
+/// Generate a consistent short hash for a MIR body.
+/// Used to avoid fn_id collisions between monomorphizations.
+pub fn hash_body(body: &Body) -> u64 {
```
cds-amal, line 291: It seems strange to hash a function and not use its mangled name, which is guaranteed to be unique. The hash seems to be structurally based and I wonder if multiple generic functions could collide. @dkcumming , what do you think?

src/mk_graph/traverse.rs
```diff
@@ -0,0 +1,194 @@
...
+                .statements
+                .iter()
+                .map(|s| ctx.render_stmt(s))
+                .collect();
+
+            let terminator = ctx.render_terminator(&block.terminator);
+
+            let cfg_edges = terminator_targets(&block.terminator)
+                .into_iter()
+                .map(|t| (t, None))
```
cds-amal: This is where information gets lost. The traversal has the full `TerminatorKind` right here; it knows whether an edge is a SwitchInt branch (with a discriminant value), a cleanup unwind, or a normal successor. But terminator_targets flattens all of that into a plain `Vec<usize>`, and then this code wraps each target in (t, None), discarding the labels entirely.

The `Option<String>` in cfg_edges exists precisely for this data. Look at what the DOT renderer needs: SwitchInt edges carry discriminant values like "0", "1", cleanup edges get "Cleanup", the otherwise branch gets "other". That information is available right now, in `block.terminator.kind`, at this exact point in the traversal. One scope exit later, it's gone; a builder that needs it would have to reach into `raw_terminator` to re-derive what the traversal already knew.

That's the pattern we're trying to break: the traversal owns the "what" (which blocks connect and what each edge means), and the builder owns the "how" (syntax, escaping, visual style). When the traversal discards edge semantics, it forces builders to do the traversal's job, and the trait boundary stops earning its keep.

I would explore matching on the terminator kind here instead of delegating to `terminator_targets`.

cds-amal (changes requested):
Thanks @0xh4ty, this has come a long way! The RenderedFunction / RenderedBlock / CallEdge structs are exactly what I had in mind, and the D2 builder is clean and easy to follow. The hard part (extracting traversal from rendering, getting the data shapes right) is done.

A few things to get us across the finish line. I left inline comments on `cfg_edges`, and `hash_body` collision; the rest I'll cover here.

raw_stmts and raw_terminator: on second thought, let's not! I want to own this one, because my earlier comment sent you in the wrong direction. I proposed `raw_terminator` as an escape hatch "for builders that need structural information; say, coloring blocks differently based on terminator kind." Having now seen how the implementation shakes out, I think we're better off without it. 

Here's the tell: `raw_stmts` and `raw_terminator` are the only reason `RenderedBlock` and `RenderedFunction` need a lifetime parameter. Remove them, and the structs become fully owned, which is cleaner for everyone. `D2Builder` already ignores both fields. And the use case I imagined (coloring based on terminator kind) doesn't need the full MIR type; a simple TerminatorTag enum (just the discriminant, no payload) would do the job if we **ever** actually need it.

The deeper issue: these fields quietly re-introduce the coupling we're trying to eliminate. The whole point of the RenderedFunction design is that builders see pre-rendered data and don't need to understand MIR internals. An escape hatch that hands them raw `&[Statement]` and `&Terminator` undermines that; it's the same boundary leak as the old `block(...)` method, just made optional instead of required. Let's take them out; if a future builder genuinely needs something the rendered data doesn't provide, that's a signal to enrich the rendered data (like the `cfg_edges` labels in my inline comment), not to punch through the abstraction.

(My apologies for the whiplash; sometimes we have to see the implementation to realize a design idea doesn't carry its weight.)

`external_function` is declared but never called: render_graph never calls it. The DOT renderer needs this: it creates red-colored nodes for functions that appear as call targets but aren't in the items list. The traversal has everything it needs to compute this. Either wire it up, or remove it from the trait until it's needed. Dead code in a trait is worse than dead code in a function, because every new impl has to carry it.

I would suggest porting DOT to the new traversal; this is the real proof that the abstraction works if DOT ports cleanly, we know the trait boundary is in the right place. If it doesn't, we'll learn something useful about what's missing. Either way, we come out ahead.

src/mk_graph/traverse.rs
0xh4ty: Got it. I’ll add matching logic on `terminator.kind` in the traversal and populate `cfg_edges` with the appropriate labels

0xh4ty:
Thanks for the clarification, that makes sense. I’ll remove `raw_stmts` and `raw_terminator` so the rendered structs become fully owned. I’ll also wire up `external_function` in the traversal to fix the dead code.

0xh4ty:
Pushed another update addressing the remaining review feedback.

Main changes:

* Removed `raw_stmts` and `raw_terminator` from the rendered graph structures, making `RenderedFunction` and `RenderedBlock` fully owned and eliminating unnecessary lifetime propagation.
* Wired up `external_function` in traversal so external call targets are computed centrally instead of by individual renderers.
* Reworked CFG edge construction to match directly on `TerminatorKind` and preserve edge semantics (switch labels, cleanup edges, etc.) instead of flattening via `terminator_targets`.
* Ported the DOT backend to the new `GraphBuilder` traversal model behind `SMIR_DOT_NEW`.
* Fixed the remaining Clippy warning by eliding an unnecessary lifetime in `render_function`.

I also generated PNG renders of the old and new DOT outputs for comparison and attached them below.

### Old DOT Implementation

fibonacci_old

### New GraphBuilder-based DOT implementation

fibonacci_new

---

Once the new DOT traversal path is confirmed to match the existing behavior, I’ll remove the legacy implementation and switch the new renderer to the default path.

0xh4ty:
Added a global `function_ids` map in `src/mk_graph/traverse.rs`.

The map is generated centrally in `render_graph` using the existing `short_name` + `hash_body` based ID scheme and is now used during traversal to resolve graph identities consistently for both functions and call edges.

While symbol names are already guaranteed to be unique by the compiler, directly embedding mangled symbols into graph node identifiers can make renderer behavior inconsistent across graph backends because many output formats impose their own parsing and identifier constraints. Using a separate hashed identity layer avoids having to escape symbol names individually for every renderer, which can make the rendered graph output harder to read and debug.

This also removes renderer-side ID reconstruction logic and gives future renderers a centralized source of graph identity, improving consistency and reliability across backends.

If a renderer prefers to use the original symbol name to reconstruct its function IDs, they can still access that value from the `symbol_name` field on `RenderedFunction`.

PR #133, opened 2026-03-02
Cachix does not respond to pin requests

F-WRunTime:
- To work around this. Introducing a script and WF update to check through API if the pin was made and the artifacts were uploaded.


dkcumming (approved):
Looks like the same solution for mir-semantics and that worked 👍

PR #134, opened 2026-03-03
Update kup publish command to allow errors

F-WRunTime:
Modify kup publish command to prevent failure on error.


dkcumming (approved):
👍 TY @F-WRunTime

PR #135, opened 2026-03-03
Factor: collapse triplicated alloc recovery into one arm

cds-amal:
This is a follow-up to #131. That PR fixed a real panic (non-builtin-deref Static/VTable allocations), but the fix landed as three nearly identical match arms: Static, VTable, and Function, each repeating the same "try `get_prov_ty`, fall back to opaque placeholder" logic. The only meaningful difference between them was a single predicate ("is this type usable directly?") and the debug log string. This PR collapses them back into one arm, makes the predicate explicit, and documents a previously unexplained jq filter.

## What changed

**`src/printer/mir_visitor.rs`**

The three arms now share a single code path. The predicate that actually differs between them is captured in a `needs_recovery` variable:

- Static/VTable: `builtin_deref(true).is_none()` (not a reference, raw pointer, or Box)
- Function: `!kind.is_fn_ptr()` (outer type isn't already a function pointer)

Worth noting: these two predicates are *not* equivalent (a function pointer passes `builtin_deref` but fails `is_fn_ptr`), which is why a naive "just combine the arms" without the inner match would have changed behavior for Function allocs. The inner match on `global_alloc` makes this asymmetry visible rather than hiding it across separate arms.

The rest of the logic (try `get_prov_ty`, fall back to opaque placeholder) is written once. Also dropped an unnecessary `.clone()` on the `builtin_deref` call; it takes `&self`.

**`tests/integration/normalise-filter.jq`**

The `def_id` filter had a terse comment ("unrelated to this regression") that didn't explain *why* it exists or why it's safe to strip globally. Replaced it with a proper explanation: `def_id` values are interned compiler indices (same class as `alloc_id` and `adt_def`) that are consistent within a single rustc invocation but non-deterministic across runs. Downstream consumers need them as cross-reference keys to join `AggregateKind::Adt` in MIR bodies with type metadata entries, so they can't be dropped from the output itself; we only strip them here for golden-file comparison. (See #125 for the full picture on interned-index non-determinism.)

## Test plan

- [ ] `cargo build` compiles cleanly
- [ ] `make integration-test` passes (no behavioral change; this is a pure refactor)

### DIFF
--- a/src/printer/mir_visitor.rs
+++ b/src/printer/mir_visitor.rs
@@ -265,12 +265,21 @@ fn collect_alloc(
                     .insert(val, (opaque_placeholder_ty(), global_alloc.clone()));
             }
         }
-        GlobalAlloc::Static(_) => {
-            // Keep builtin-deref behavior; recover only non-builtin-deref cases.
-            if kind.clone().builtin_deref(true).is_none() {
+        GlobalAlloc::Static(_) | GlobalAlloc::VTable(_, _) | GlobalAlloc::Function(_) => {
+            // Does the outer type need provenance recovery to find the real
+            // pointee? Static/VTable pointers are usable directly when
+            // builtin_deref succeeds; Function allocs are usable directly
+            // when the outer type is already a fn pointer.
+            let needs_recovery = match &global_alloc {
+                GlobalAlloc::Function(_) => !kind.is_fn_ptr(),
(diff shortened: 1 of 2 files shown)


dkcumming (approved):
Nice! Tested downstream with mir-semantics (682df6b2dfec4ec0516e246f22b70dff09cf57a6) test suite which is passing

PR #142, opened 2026-03-10
Shake your foundation: make, don't panic & derive..

cds-amal:
This PR introduces independent improvements that lays the groundwork for robust multi-nightly
support. Each one stands alone, but together they establish the tooling and
resilience the rest of the stack builds on.

- feat(make): self-documenting help target and modernized Makefile. `make help`
  now extracts target descriptions via an awk one-liner, instead of complicated
  make shellery. Also adds standalone `fmt`, `clippy`, `stdlib-smir`, and `build-info`
  targets. The Makefile establishes its final structure here (including stubs
  for golden file management and UI testing that become functional in later
  PRs); we write the API we desire and make it real in subsequent PRs.

- fix(ty_visitor): catch `rustc` panics during layout computation. Some types
  (notably `dyn` `Trait` in certain positions) cause rustc's layout engine to panic
  rather than returning Err. Turns out `catch_unwind` is the only viable option
  here; rustc doesn't surface these as recoverable errors. We wrap `ty.layout()`,
  record panicked types, and report them in a summary instead of taking the
  whole run down with them.

- fix(cargo_stable_mir_json): derive toolchain lib path dynamically. The helper
  binary hardcoded the library path, which broke every time the nightly
  directory name changed (i.e., on every nightly bump; not ideal). Now resolves
  it at runtime from the active toolchain.

- refactor: derive rustc commit from toolchain, drop [metadata] section.
  rust-toolchain.toml carried a [metadata] rustc-commit = ... field that had to
  be updated by hand on every bump. The UI test scripts now derive the commit
  hash from the Rust release manifest instead, which means one fewer thing to
  forget, yay! Also inlines format args across the codebase to satisfy
  `clippy::uninlined_format_args`.

Test plan

- make help prints a formatted target list
- cargo build succeeds on the pinned nightly
- make clippy and make fmt work as standalone targets
- `cargo run -- -Zno-codegen tests/integration/programs/enum.rs` produces
output (exercises the `ty_visitor` panic catch on types with tricky layouts)

### DIFF
--- a/.github/scripts/install-yq.sh
+++ b/.github/scripts/install-yq.sh
@@ -12,13 +12,19 @@ BASE_URL="https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}"
 mkdir -p "$INSTALL_DIR"
 
 wget -qO "$INSTALL_DIR/yq_linux_amd64" "${BASE_URL}/yq_linux_amd64"
-wget -qO "$INSTALL_DIR/checksums"       "${BASE_URL}/checksums"
+wget -qO "$INSTALL_DIR/checksums-bsd"   "${BASE_URL}/checksums-bsd"
 
 (
     cd "$INSTALL_DIR"
-    grep 'yq_linux_amd64$' checksums | sha256sum -c -
+    # checksums-bsd uses BSD-style:   SHA256 (yq_linux_amd64) = 0c4d965e...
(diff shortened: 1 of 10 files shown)


dkcumming (approved):
Awesome stuff! The commits related to the panic unwind especially were really interesting! Sorry for the delay on getting this reviewed. I ran the mir-semantics test suite with this as a dep and no regressions as expected too.

PR #144, opened 2026-03-11
Bedrock, Fault lines and Striking Gold

cds-amal:
So far, the integration test golden files live next to the test programs (tests/integration/programs/*.expected), and the normalise filter uses hardcoded per-field rules to strip interned indices. This works fine when you only support one nightly. It stops working the moment you need to support multiple nightlies in parallel: each nightly may produce structurally different MIR output (new enum variants, removed fields, renamed types), so a single set of golden files can't serve as the expected output for all of them.

This PR introduces the infrastructure to solve that problem, in four parts:

- build.rs breakpoint detection: parses rustc -vV's commit-date and emits cargo:rustc-cfg flags for 16 known API breakpoints. The compat layer uses these flags to gate match arms and import paths, so the same codebase compiles against any nightly from 2024-11-29 through 2026-01-15. The breakpoints table is date-sorted, with a naming convention (smir_has_<thing> for additions, smir_no_<thing> for removals) that makes it straightforward to extend.

- Per-nightly golden file directories: expected outputs move from flat tests/integration/programs/*.expected to tests/integration/expected/<nightly>/. The Makefile's integration-test target auto-detects the active nightly from rustc -vV and selects the matching golden directory, falling back to the pinned nightly if no directory exists for the active one.

- Receipt-driven normalise filter: the jq filter now reads the companion *.smir.receipts.json (from PR #143) and applies interned-index normalization generically. No more hardcoded per-field rules; the filter walks the JSON tree and strips any key, newtype, or array position that the receipts declare as interned. This is where the receipts system pays off: new interned fields are handled automatically without filter updates.

- nightly_admin.py: a Python script (with Make targets nightly-add, nightly-check, nightly-bump) that formalizes the manual workflow for absorbing a new nightly. add installs the toolchain, generates golden files, and emits UI test lists; check runs integration + UI tests against a specific nightly; bump updates rust-toolchain.toml and README.md. What used to be a sequence of a dozen shell commands and file edits is now a single invocation.

This PR also includes the compat code for the first five breakpoints (nightly-2024-12-14 through nightly-2025-03-01) and the initial set of receipt-normalized golden files for nightly-2025-03-01. The subsequent nightly PRs in this stack each add one breakpoint and its corresponding golden files; they're small and mechanical by design, because all the infrastructure is here.

Test plan

- cargo build passes (breakpoint detection should activate the first five flags)
- cargo clippy -- -Dwarnings passes
- make integration-test passes (auto-selects tests/integration/expected/nightly-2025-03-01/ golden files with receipt-driven normalization)
- make golden regenerates golden files into the correct per-nightly directory


scripts/nightly_admin.py
```diff
@@ -0,0 +1,472 @@
...
+    """Return env dict that forces a specific toolchain."""
+    return {"RUSTUP_TOOLCHAIN": nightly}
+
+
+def read_file(path: Path) -> str:
+    return path.read_text(encoding="utf-8")
+
+
+def write_file(path: Path, content: str) -> None:
+    path.write_text(content, encoding="utf-8")
```
dkcumming: No guard on failure

scripts/nightly_admin.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+def count_lines(path: Path) -> int:
+    """Count non-empty, non-comment lines in a TSV file."""
+    if not path.exists():
+        return 0
```
dkcumming: Feels like this should not be zero and should be an error?

scripts/nightly_admin.py
```diff
@@ -0,0 +1,472 @@
...
+# ---------------------------------------------------------------------------
+# DEFAULT_NIGHTLIES management
+# ---------------------------------------------------------------------------
+
+def read_default_nightlies() -> list[str]:
+    """Parse DEFAULT_NIGHTLIES from diff_test_lists.sh."""
+    content = read_file(DIFF_SCRIPT)
+    m = re.search(r'DEFAULT_NIGHTLIES="([^"]*)"', content)
+    if not m:
+        die("could not parse DEFAULT_NIGHTLIES from diff_test_lists.sh")
```
dkcumming: This is referencing a script that is not in this PR, I looked ahead and it is in a future one

scripts/nightly_admin.py
```diff
@@ -0,0 +1,472 @@
...
+from pathlib import Path
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+RUST_TOOLCHAIN_TOML = REPO_ROOT / "rust-toolchain.toml"
+README_MD = REPO_ROOT / "README.md"
+DIFF_SCRIPT = REPO_ROOT / "tests" / "ui" / "diff_test_lists.sh"
```
dkcumming: This is referencing a file that does not exist in this repo during this commit.

scripts/nightly_admin.py
```diff
@@ -0,0 +1,472 @@
...
+GOLDEN_BASE = REPO_ROOT / "tests" / "integration" / "expected"
+OVERRIDE_BASE = REPO_ROOT / "tests" / "ui" / "overrides"
```
dkcumming: This also doesn't exist

scripts/nightly_admin.py
```diff
@@ -0,0 +1,472 @@
...
+    return m.group(1).split()
+
+
+def write_default_nightlies(nightlies: list[str]) -> None:
+    """Rewrite the DEFAULT_NIGHTLIES line, keeping sorted order."""
+    content = read_file(DIFF_SCRIPT)
+    sorted_list = sorted(nightlies)
+    new_line = f'DEFAULT_NIGHTLIES="{" ".join(sorted_list)}"'
+    updated = re.sub(r'DEFAULT_NIGHTLIES="[^"]*"', new_line, content)
+    write_file(DIFF_SCRIPT, updated)
```
dkcumming: This also only makes sense in context of future content.

PR #145, opened 2026-03-11
Read the room, skip the rest

cds-amal:
The UI tests (where we run stable-mir-json against rustc's own test suite) have a versioning problem that mirrors the golden file problem solved in PR #144: different nightlies have different sets of UI tests. Files get added, deleted, and renamed between nightly commits. A test that exists in nightly-2025-03-01 might be gone by nightly-2025-10-03, or moved to a different directory. Running a stale test list against a newer nightly produces spurious failures; maintaining the lists by hand is tedious and error-prone.

This PR adds the tooling to generate, validate, and run per-nightly UI test lists automatically. Three pieces:

- parse_test_directives.awk: an awk script that extracts `//@` directives from rustc test source files and decides whether a test should be skipped on the current host. It handles `only-<target>`, `ignore-<target>`, `needs-sanitizer`, `needs-subprocess-spawning`, edition directives, `compile-flags`, and a handful of environment-specific skips. A "universal" mode suppresses platform-specific filtering so that generated lists are correct on any host; platform filtering happens at runtime instead.

> One environment-specific skip worth calling out: tests that reference `extern crate libc` are skipped because our sysroot contains both `.rmeta` and `.rlib` artifacts for libc. Rustc, invoked directly outside cargo, sees two candidates for the same crate and bails with E0464. Cargo normally sidesteps this by passing `--extern libc=/exact/path`, but we don't have that luxury. The skip is a pragmatic workaround, not a judgment on the tests themselves.

- diff_test_lists.sh: given a rust-lang/rust checkout, this script diffs the tests/ui/ directory between the base nightly commit and a target nightly commit. It tracks file deletions, renames, and additions, then applies them to the base passing.tsv/failing.tsv to produce effective per-nightly test lists. The output is deterministic: same repo + same commits = same lists. Supports --report (human-readable diff summary), --emit (write lists to tests/ui/overrides/<nightly>/), and --chain (show incremental diffs between consecutive nightlies).

- Rewrites of run_ui_tests.sh and remake_ui_tests.sh: both now use the shared directive parser instead of inline awk snippets, pick up per-nightly override lists when available, and handle architecture filtering correctly. run_ui_tests.sh also fixes the RUN_SMIR library path issue that caused failures on some setups.

The PR includes pre-generated override lists for all 13 supported nightlies (2025-03-01 through 2026-01-15), a unit test suite for the directive parser (test_directives_test.sh, ~420 lines of boundary-condition tests), and corresponding Makefile targets (make test-ui, make test-ui-emit, make test-directives).

## Test plan

- make test-directives passes (unit tests for the awk parser)
- make test-ui RUST_DIR_ROOT=/path/to/rust passes with the pinned nightly
- make test-ui-emit RUST_DIR_ROOT=/path/to/rust NIGHTLY=nightly-2025-03-01 generates lists matching the checked-in overrides


tests/ui/parse_test_directives.awk
```diff
@@ -0,0 +1,145 @@
...
+#
+# The FLAGS line may have an empty flags string if no compile-flags were found.
+
+BEGIN {
+    skip = ""
+    flags = ""
+
+    # Map host_os to the set of OS names this host satisfies.
+    # "unix" covers linux, macos, freebsd, etc.  "apple" covers macos.
+    is_unix  = (host_os == "linux" || host_os == "macos" || host_os == "freebsd" || host_os == "openbsd" || host_os == "netbsd" || host_os == "dragonfly" || host_os == "solaris" || host_os == "illumos" || host_os == "android")
```
cds-amal, line 28: I'm open to paring this down :) Solaris, it's been a while.

cds-amal (commented):
A note to reviewers, pay attention to the `parse_test_directives.awk`.

PR #150, opened 2026-07-27
Enable CI, disable CD

ehildenb:
This removes all uses of secrets from this repository CI, which makes it safe to run again.

This also removes:

- Automatic version bumps on every PR.
- Cutting releases on master pushes.
- Deploying dockerhub containers with stable-mir-json installed.
- Deploying nix caches with stable-mir-json.
- Updates the versions of all pulled-in actions to the latest available.
- Move all workflows to the public github runners except the nix ones.
- RV cachix accessors and uses public access to nix instead.

### Observations on runners
[Logs](https://github.com/runtimeverification/stable-mir-json/actions/runs/30578697491/job/90993563961?pr=150) indicate that the nix jobs are being run on the self-hosted runners. These runners have Nix preinstalled and so will skip the install steps.  Therefore `install_url` and `extra_nix_config` do nothing, and the job runs on the local machine's nix ( evidently 2.28.3) rather than the pinned 2.30.1. There are some failing attempts to pull state from old cachix tokens that are not breaking the tests. If the state gets scrubbed, the lower timeout (~10) might be a problem as the self hosted runners could easily exceed this in just the build step alone.


dkcumming (commented):
The nix job still relies on cachix access to K, so that will either need to be stripped or solved.

Only other comment is since so much is going in this PR, is there are point keeping `deps/rv-nix-tools`, I don't think it gets read or written after this change right? 

EDIT: I looked into removing the `deps/rv-nix-tools` and I was not sure exactly how to change the `flake.nix`, but since it is reading from github I think it still works fine so I am not touching it.

cds-amal (approved):
Hi @dkcumming , 

Nothing untoward pops out. The only concern is the Claude::opus comment on test.yml 

>  Nix Tests is left alone deliberately. Its matrix includes the MacM1
    self-hosted host, and it skips the Nix and Cachix install steps there
    because that host has Nix preinstalled; retargeting it means dropping
    those guards and installing Nix on a GitHub-hosted macOS runner. More
    importantly, dropping deployment CI removed the job that populated the
    k-framework cache with this repo's builds, so 'nix flake check' now
    rebuilds cold. The self-hosted runners still have a persistent /nix/store
    to absorb that; a GitHub-hosted runner would not, and the job's
    timeout-minutes: 20 would very likely be exceeded.

I think the potentially time-sink consequence should be documented in the PR or README so future you won't wonder why. 

I'm also approving .. and leaving anything **nix related** to your better judgement. :)

dkcumming (approved):
Approving but I think we should iron out what is going on the self hosted runners at some stage

